PR #8, opened 2021-02-18
Feature/historic data to json

keebrev:


data/datamodule.py
```diff
@@ -58,25 +72,61 @@ def load_exchange(self):
...
+        """
         for pair in self.config['pairs']:
-            self.history_data[pair] = []
-            self.download_data_for_pair(pair)
+            if not self.check_for_datafile_existence(pair, self.config['timeframe']):
+                # datafile doesn't exist. Start downloading data, create datafile
+                print("[INFO] Did not find datafile for %s" % pair)
+                self.history_data[pair] = []
+                self.download_data_for_pair(pair, True)
+                continue
```
marijn111, line 93: Continue is redundant here. if this 'if' condition triggers, it won't evaluate the other 'elifs'.

data/datamodule.py
```diff
@@ -58,25 +72,61 @@ def load_exchange(self):
...
+            elif self.does_datafile_cover_backtesting_period(pair, self.config['timeframe']):
+                # load data from datafile instead of exchange
+                self.history_data[pair] = []
+                self.read_data_from_datafile(pair, self.config['timeframe'])
+            elif not self.does_datafile_cover_backtesting_period(pair, self.config['timeframe']):
+                # remove file, download data from exchange, create new datafile
+                self.history_data[pair] = []
+                self.delete_file(pair, self.config['timeframe'])
+                self.download_data_for_pair(pair, True)
```
marijn111, line 102: Eventually, This should append to existing data if its there. 

is there data? -> is it for the right timeframe? -> is it for the right pair? -> does it cover the specified period? -> fill in only the missing candles if needed

Current implementation suffices for now.

data/datamodule.py
```diff
@@ -103,18 +164,169 @@ def config_timeframe_calc(self):
...
+        """
+        :param pair: Certain coin pair in "AAA/BBB" format
+        :type pair: string
+        :param timeframe: Time frame of coin pair f.e. "1h" / "5m"
+        :type timeframe: string
+        :return: Returns whether datafile for specified pair / timeframe already exists
+        :rtype: boolean
+        """
+        filename = self.generate_datafile_name(pair, timeframe)
+        dirpath = "data/backtesting-data/" + self.config["exchange"] + "/" + filename
```
marijn111, line 201: Pathnames should be generated (dynamically) with os.path.join. Saves headaches in the future.

data/datamodule.py
```diff
@@ -103,18 +164,169 @@ def config_timeframe_calc(self):
...
+
+        if not path.exists("data/backtesting-data/%s" % self.config["exchange"]):
+            try:
+                os.makedirs("data/backtesting-data/%s" % self.config["exchange"])
+            except OSError:
+                print("Creation of the directory %s failed" % path)
+            else:
+                print("Successfully created the directory %s " % path)
+                return False
```
marijn111, line 210: Creating the directory if its not there should be a separate function. This looks confusing here, since the function checks for existence and returns bool based on that. If you would look at the places where this function is called, you would have no way of knowing whether the directory is being made or not.

data/datamodule.py
```diff
@@ -103,18 +164,169 @@ def config_timeframe_calc(self):
...
+        :rtype: None
+        """
+        filename = self.generate_datafile_name(pair, timeframe)
+        filepath = "data/backtesting-data/" + self.config["exchange"] + "/" + filename
+        try:
+            with open(filepath, 'r') as datafile:
+                data = datafile.read()
+        except FileNotFoundError:
+            print("[ERROR] Backtesting datafile was not found.")
+            raise SystemExit
```
marijn111, line 233: I think Demian already somewhat explained this, but I'll elaborate a bit more here. My concern is that this type of error handling is not debug friendly. Lets say this triggers. the print is shown. Then, the sysexit is raised, which in practice is a stack trace of 30+ lines. Good luck finding the print messages above all that. Furthermore, print goes to standard out, but the exception text goes to standard error (which is probably what we want).

Ideally something like this:
```python
except FileNotFoundError:
    # maybe even send it to a logger here first, and then:
    raise FileNotFoundError("[ERROR] Backtesting datafile was not found.")
```

Should be tested a little bit what works best and adopt it project wide.

data/datamodule.py
```diff
@@ -103,18 +164,169 @@ def config_timeframe_calc(self):
...
+            print("[ERROR] Backtesting datafile was not found.")
+            return False
+        except:
+            print("[ERROR] Something went wrong loading datafile", sys.exc_info()[0])
+            return False
+
+        historic_data = json.loads(data)
+        if historic_data["from"] == self.backtesting_from and historic_data["to"] == self.backtesting_to:
+            return True
+        return False
```
marijn111, line 321: Should be fairly simple to make this a datetime range, and check if the specified dates falls in between the existing. That should also make it relatively easy to extract the days that are missing, and implement some logic that only downloads the missing days

data/datamodule.py
```diff
@@ -103,18 +164,169 @@ def config_timeframe_calc(self):
...
+    def customOHLCVDecoder(self, ohlcv_dict) -> namedtuple:
+        """
+        This method is used for reading ohlcv-data from saved json datafiles.
+
+        :param ohlcv_dict: dictionary-format ohlcv-model, which is 1 candle in specified timeframe
+        :type ohlcv_dict: json-format ohlcv-model
+        :return: named tuple with OHLCV properties (more or less the same as the model)
+        :rtype: namedtuple
+        """
+        return namedtuple('OHLCV', ohlcv_dict.keys())(*ohlcv_dict.values())
```
marijn111, line 332: Nice! +1 for approaching it like this

marijn111 (approved):

backtesting/backtesting.py
```diff
@@ -228,3 +270,32 @@ def calculate_max_seen_drawdown(self):
...
+        """
+        Helper method for initializing defaultdict containing arrays
+        """
+        return []
+
+    def default_empty_dict_dict(self):
+        """
+        Helper method for initializing defaultdict
+        """
+        return defaultdict()
```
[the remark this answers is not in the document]
dema-trading-ai: @marijn111 I've used this because the created dictionary contains dictionaries

dema-trading-ai (approved):

Issue #11, opened 2021-02-18
Dockerize engine to prevent dependency errors


dema-trading-ai:
We got some messages that people had issues with dependencies as CCXT and TA-Lib. For this reason, it is nice to dockerize the engine.

Things needed:
- TA-Lib binaries install script
- Dockerfile
- Docker-compose file

PR #12, opened 2021-02-19
Feature/dockerize #11

dema-trading-ai:
We fixed issue #11 in this pull-request. To make this work Tijs added:

- Install scripts for TA-Lib binaries
- TA-Lib binaries as installing helpers
- Dockerfile
- Updated README for using Docker


noudcorten (approved):
Checked all files and everything looked fine.

pvdklei:
Did not try docker yet, but the ta-lib installation bash script works. I only still get this `Symbol not found: _TA_ACOS` error when doing `import talib`, but that also occurred when installing earlier. Has something to do with my newer laptop, it does work on my older one.

backtesting/backtesting.py
```diff
@@ -54,8 +54,8 @@ def start_backtesting(self, data, backtesting_from, backtesting_to) -> None:
             for tick in data[pair]:
                 data_per_tick[tick.time].append(tick)
 
+        ticks_passed_per_pair = defaultdict(self.default_empty_array_dict)
```
DemianVoorhagen6, line 57: Moving this means it won't be re-made every time, I assume that is why it is moved up?

data/datamodule.py
```diff
@@ -196,8 +196,8 @@ def check_for_datafile_existence(self, pair, timeframe) -> bool:
         :return: Returns whether datafile for specified pair / timeframe already exists
         :rtype: boolean
         """
-        filename = self.generate_datafile_name(pair, timeframe)
-        dirpath = os.path.join("data/backtesting-data", self.config["exchange"], filename)
+        pair_without_slash = pair.split('/')
+        dirpath = os.path.join("data/backtesting-data", self.config["exchange"], pair_without_slash[0], pair_without_slash[1])
```
DemianVoorhagen6, line 200: I'd argue that the following approach would be slicker:
coin, base = pair.split('/')
dirpath = os.path.join("data/backtesting-data", self.config["exchange"], coin, base)

data/tradingmodule.py
```diff
@@ -293,14 +293,12 @@ def update_drawdowns_closed_trade(self, trade: Trade) -> None:
         if trade.profit_percentage < 0:
             if self.last_closed_trade is None:
                 self.temp_realized_drawdown = trade.profit_percentage
-                self.last_closed_trade = trade
-                return
-            if self.last_closed_trade.profit_percentage < 0:
+            elif self.last_closed_trade.profit_percentage < 0:
                 self.temp_realized_drawdown += trade.profit_percentage
```
DemianVoorhagen6, line 297: Sorry I don't have time to check atm, but can you please check if this logic holds? If for example you have 100 euro start capital, you lost 20 euro - you are at 20% drawdown now. Now lets say you open a trade with 80 euro, and you lose another 20 euro. This would be 25% drawdown on the trade (80 > 60), but  40% on the total portfolio (100 > 60)

data/tradingmodule.py
```diff
@@ -293,14 +293,12 @@ def update_drawdowns_closed_trade(self, trade: Trade) -> None:
...
-            self.temp_realized_drawdown = trade.profit_percentage
+            else:
+                self.temp_realized_drawdown = trade.profit_percentage
```
DemianVoorhagen6, line 299: Does this imply that is you make a profitable trade you are out of drawdown? Because thats not the case either. If you are at 50% drawdown, and make a 1% winning trade, you still have 49,5% drawdown.

data/tradingmodule.py
DemianVoorhagen6, line 297: Sorry I don't have time to check atm, but can you please check if this logic holds? If for example you have 100 euro start capital, you lost 20 euro - you are at 20% drawdown now. Now lets say you open a trade with 80 euro, and you lose another 20 euro. This would be 25% drawdown on the trade (80 > 60), but 40% on the total portfolio (100 > 60)

data/datamodule.py
DemianVoorhagen6, line 200: I'd argue that the following approach would be slicker:
coin, base = pair.split('/')
dirpath = os.path.join("data/backtesting-data", self.config["exchange"], coin, base)

backtesting/backtesting.py
DemianVoorhagen6, line 57: Moving this means it won't be re-made every time, I assume that is why it is moved up?

noudcorten: Yes. Moving it up will result in the past ticks actually being stored and not overwritten at every iteration.

data/datamodule.py
noudcorten: Agreed.

backtesting/backtesting.py
DemianVoorhagen6: yes I got that, but they both mean different things, so we have to make sure this is wanted behaviour :)

data/tradingmodule.py
keebrev: This is a good point. Will have a look on that.

backtesting/backtesting.py
noudcorten: The goal of this dict is to store the seen ticks per pair which with this update will be its behaviour. If you look at the previous code closely, the input for line 60 will always be: (current tick, empty list). This is not what we want.

data/tradingmodule.py
keebrev: True. Fixed that in next commit

data/datamodule.py
keebrev: Fixed in next commit

PR #25, opened 2021-02-24
Better pip caching when building images 

pvdklei:
After this change docker will only reinstall dependencies from `requirements.txt` whenever this file is changed. Before, it would run it every time, which results in extra build time


dema-trading-ai:
Please re-open this PR to 'development' branche, as Dockerize is merged already

keebrev:
How many MB does this differ from the old method?

dema-trading-ai (approved):

marijn111 (approved):

PR #29, opened 2021-02-24
Feature/config base curr

### LINKED ISSUE #6: Create configurable base currency
In the current state, we can add any coin pair we want. Even if the base currency differs. For example, we can add both BTC/USDT and ETH/BTC to our pairlist. From this point, calculated profit won't be accurate at all. 

We need next functionality:
- Base currency configurable through config file
- Check pairlist for validity (if base currency is USDT, only pairs ending with "***/USDT" would be valid
- Change currency signs in backtest result based on specified currency (currently everything is displayed with $ sign)

pvdklei:
In this PR I fixed issue #6. I added a 'currency' entry to the config file, and validated it. For instance checking whether this entry exists, and whether it it the same as the currency in the pairs. Finally, I changes the currency symbols shown in the output (only for USD, EUR, GDB and JPY for now, but more can be added easily). 

In addition I moved config validation to a separate module. 

Closes #6


config/validations.py
```diff
@@ -0,0 +1,16 @@
...
+    pairs = config["pairs"]
+    currency = config["currency"]
+    for pair in pairs:
+        pair = pair.split("/")
+        assert len(pair) == 2
+        if not pair[1] == currency:
+            print(
+                "[ERROR] You can only use pairs that have the base currency you specified")
+            print(
+                "[ERROR] e.g., if you specified 'USDT' as your currency, you cannot add 'BTC/EUR' as a pair")
```
dema-trading-ai, line 15: Nice!

config/checks.py
```diff
@@ -0,0 +1,34 @@
...
+    if config_json.get('max-open-trades') is None:
+        print("[ERROR] You should define max_open_trades in the config-file")
+        raise SystemExit
+
+
+def check_config_base_currency(config_json):
+    if config_json.get("currency") is None:
+        print(
+            "[ERROR] You should define your base currency as 'currency' in the config-file")
+        raise SystemExit
```
dema-trading-ai, line 34: Could you have a look at #30 and implement Marijns' approach of checking in here?

dema-trading-ai (changes requested):

keebrev (approved):

dema-trading-ai (approved):

PR #30, opened 2021-02-24
Fixes #28 - Checking presence of required params in config

### LINKED ISSUE #28: Optimize checking presence of required config params
The current approach is not scalable and will soon prove unmanageable if more and more (required) config params are added.

https://github.com/dema-trading-ai/engine/blob/0bd655d12deaaa0f82e574f3178ffb6f74c83ede/main.py#L26-L29

Make one function that takes the required param as argument and check whether it is present in the config

Furthermore, the following check doesn't work as expected, and will throw a keyerror if not present (instead of the desired handling)
https://github.com/dema-trading-ai/engine/blob/0bd655d12deaaa0f82e574f3178ffb6f74c83ede/main.py#L42

Something like this should do the trick
```python
if 'stoploss' not in config_json: 
```

marijn111:
# Fixes

Closes #28 

# What has been changed?

Checking the config for any required params is optimised. One function that checks for presence, instead of a separate function for each param.

# Examples

N/A

Depending on #12 

# Security Measures

N/A

Instantiating the required params could probably be optimised a bit in the future. (e.g. extend it with extra checks for each param. Think of the stoploss has to be between 0 and 100)

# Additional Info

PR Depends on #29

### DIFF
--- a/main.py
+++ b/main.py
@@ -3,6 +3,13 @@ import sys
 
 from main_controller import MainController
 
+REQUIRED_PARAMS = [
+    'stoploss',
+    'max-open-trades',
+    'fee',
+    'starting-capital',
+]
+
 
 def main():
     read_config()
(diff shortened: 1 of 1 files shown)


dema-trading-ai:
@marijn111 this is already implemented by Pepijn!

Issue #17, opened 2021-02-22
Review / Update backtest result overview


keebrev:
On the backtest-overview, a couple of things needs to be changed. 

Things to be revised:
- General 'generate backtest result' method. This method is way to large
- Tabs / indents in docker result 

Things to be added to the backtest overview:
- Total fee paid
- Configured stoploss
- Put 'seen' in front of 'drawdown from' / 'drawdown to'
- From - To on 1 line
- Ended with $ / % on 1 line

PR #53, opened 2021-03-02
Review/backtest results overview

pvdklei:
As #17 suggested that `BackTesting.generate_backtesting_results()` was too large, and could use some cleanup, I refactored this function. I did this by creating some new result-models: `MainResults`, `CoinInsights` and `OpenTradeResult` in `results.py`, that handle their own visualisation. By removing the `print` statements to other classes, `BackTesting` can fully focus on only calculating these results. 

A next step would be to use these models in all places and earlier on in calculations, as the values that these models own are currently stored temporarily in `dict`'s and other variables (unnecessary memory transfers) in the methods that create these results in `BackTesting`.

Note: this PR does not fully solve #17.


dema-trading-ai (approved):

dema-trading-ai (commented):
What is missing from #17 ?

pvdklei:
Actually just the first bullet point in #17 is solved: "General 'generate backtest result' method. This method is way to large"

pvdklei:
#17 is now fully solved, apart from putting "from - to" on one line (that looked odd, and made the table too wide) and adding total fee paid (fee calculations must first be implemented).

keebrev:
> #17 is now fully solved, apart from putting "from - to" on one line (that looked odd, and made the table too wide) and adding total fee paid (fee calculations must first be implemented).

Very nice! Alright, if from - to looks odd, let's skip it. It was just an idea to improve the results visually.

makefile
```diff
@@ -0,0 +1,10 @@
+all: build run
+
+build: 
+	docker build . -t dema-engine:alpha
+
+run:
+	docker run --rm dema-engine:alpha
+
+runv: 
+	docker run --rm -v "$(shell pwd):/engine" dema-engine:alpha
```
keebrev, line 10: Really like this! But could you please add to README how we can use this and what is needed for this to work? (I assume we need some binary package)

keebrev (changes requested):

backtesting/results.py
```diff
@@ -0,0 +1,119 @@
...
+        print("| Backtesting from: \t\t%s" % self.tested_from)
+        print("| Backtesting to: \t\t%s" % self.tested_to)
+        print("| ")
+        print("| Started with: \t\t%s" % round(self.starting_capital, 2) + '\t' + currency_symbol)
+        print("| Ended with: \t\t\t%s" % round(self.end_capital, 2) + '\t' + currency_symbol)
+        print("| Overall profit: \t\t%s" % round(self.overall_profit_percentage, 2) + '\t%')
+        print("| Amount of trades: \t\t%s" % self.n_trades)
+        print("| Left-open trades: \t\t%s" % self.n_left_open_trades)
+        print("| Trades with loss: \t\t%s" % self.n_trades_with_loss)
+        print("| Stoploss: \t\t\t%s" % self.configured_stoploss + "\t" + currency_symbol)
```
keebrev, line 48: Stoploss is not in the base currency but in percentages. Could you please fix? :)

makefile
pvdklei: Accidentally left it in, but nice that you like it!! You need `make` installed, but I think it's installed by default on most computers. I'll add it.

backtesting/results.py
pvdklei: Sure!

pvdklei:
Changed the $ to a % and changed the README. Can someone on windows and linux maybe check if `make` is installed by default? Thought so, but can't get confirmation online or some good one-liners to install it.

keebrev:
> Changed the $ to a % and changed the README. Can someone on windows and linux maybe check if `make` is installed by default? Thought so, but can't get confirmation online or some good one-liners to install it.

I am not sure for both the OS's. As every environment will differ, I don't think you can assume that it's installed. However, it's only 1 simple step installing / checking whether it's installed.

Checking whether it's installed
```
make --version
```

Installing on Linux
```
 sudo apt install make
```

Installing on Windows (chocolately required)
```
choco install make
```

pvdklei:
True, so I've added it. To not bloat the README too much, I've put the `make` install instructions in-line with the rest.

keebrev (approved):

PR #70, opened 2021-03-06
Adding exchange fees to trading module and config

sid-chaubs:


noudcorten:
You are now merging this request into the 'main' branch. This needs to be changed to the 'development' branch.

data/tradingmodule.py
```diff
@@ -139,14 +139,15 @@ def open_trade(self, ohlcv: OHLCV) -> None:
...
         available_spaces = self.max_open_trades - open_trades
         amount = ((100 / available_spaces) / 100) * self.budget
-        self.budget-=amount
+        self.budget -= amount
         new_trade = Trade()
         new_trade.pair = ohlcv.pair
         new_trade.open = ohlcv.close
         new_trade.current = ohlcv.close
         new_trade.status = "open"
         new_trade.amount = (amount / ohlcv.close)
```
noudcorten, line 148: Fee needs to be issued here as well. Budget calculation is correct, but for the 'trade amount' an extra line needs to be added:
`trade_amount = amount - (amount * float(self.config['exchange_fees'])`

Maybe as an extra optimisation. Add line to __init__():
`self.fee = float(self.config['exchange_fees'])`

data/tradingmodule.py
```diff
@@ -232,8 +235,19 @@ def update_opentrade_stats(self, ohlcv: OHLCV) -> Trade:
         """
         trade = self.find_open_trade_for_pair(ohlcv.pair)
         trade.current = ohlcv.close
-        trade.profit_percentage = ((ohlcv.close - trade.open) / trade.open) * 100
-        trade.profit_dollar = (trade.amount * trade.current) - (trade.amount * trade.open)
+
+        # multiplying by 2 in order to account for opening and closing the trade
+        total_fees_percent = trade.exchange_fees * 2
```
pvdklei, line 240: Fees get paid when buying AND selling right? So lets say fee is 1% and you buy for 100$, so first fee is 1$. Now price increases to 120$, and you decide to sell. This time fee is 1.20$, so the total fee of the trade is 2.20$. If you take this as a percentage over the starting price it is 2.20%, right? Not the 2 x 1% = 2% that you calculate here. In almost every occasion you can't just add, subtract and multiply raw percentages...

models/trade.py
```diff
@@ -19,3 +19,4 @@ class Trade:
     sell_reason = None
     opened_at = None
     closed_at = None
+    exchange_fees = None
```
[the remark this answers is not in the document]
pvdklei: True

data/tradingmodule.py
noudcorten: Exactly. That's why it needs to be changed here https://github.com/dema-trading-ai/engine/pull/70#discussion_r590178041

noudcorten:
@dema-trading-ai This pull request is already handled in https://github.com/dema-trading-ai/engine/pull/107 right?

sid-chaubs:
> @dema-trading-ai This pull request is already handled in #107 right?

@noudcorten closing this pull request since the changes are handled in #107 and merged in. Thanks for bringing this to my attention!

PR #71, opened 2021-03-08
cleanup calculation

### LINKED ISSUE #48: Check profit calculations properly
I have some doubts about all my calculations and therefore want it to be checked by somebody else. 

Things to check:
- Overall profit calculations (both %/$)
- Sell Reason count methods (ROI/SL/Signal)
- Trade open/close method calculations
- Per coin statistics

Things to do:
- Make all the calculations more readable
- Make separate methods for calculations (if calculation is redundant)
- Optimize calculations

deborahvans:
# Fixes
Fixes #48

# What has been changed?
Checked calculation; ran and recalculated everything
Removed redundant sell reasons dict (from generate coin result)
renamed var for avg_profit_prct
moved helper functions to utils

# Examples
As it was, the sell seasons dict was first calculated and then recalculated in exacly the same format.

None

# Security Measures
n/a

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -7,15 +7,12 @@ def get_project_root():
 
 
 def default_empty_array_dict() -> list:
-    """ :return: list for initializing dictionary :rtype: List """
+    """:return: list for initializing dictionary :rtype: List"""
     return []
 
 
 def default_empty_dict_dict() -> dict:
-    """
-    :return: Dictionary for initializing default dictionary
-    :rtype: Dict
-    """
+    """:return: Dictionary for initializing default dictionary :rtype: Dict"""
     return defaultdict(int)
 
 
--- a/utils.py
+++ b/utils.py
@@ -1,15 +1,15 @@
 from pathlib import Path
 from collections import defaultdict
 
+
 def get_project_root():
     return Path(__file__).parent
 
+
 def default_empty_array_dict() -> list:
-    """
-    :return: list for initializing dictionary
-    :rtype: List
-    """
+    """ :return: list for initializing dictionary :rtype: List """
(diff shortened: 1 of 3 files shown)


utils.py
```diff
@@ -1,4 +1,31 @@
 from pathlib import Path
+from collections import defaultdict, namedtuple
 
 def get_project_root():
     return Path(__file__).parent
+
+def default_empty_array_dict() -> list:
+    """
+    Helper method for initializing defaultdict containing arrays
+    """
```
noudcorten, line 10: add :param: and :return: to comments

utils.py
```diff
@@ -1,4 +1,31 @@
...
+    return []
+
+def default_empty_dict_dict() -> dict:
+    """
+    Helper method for initializing defaultdict
+    """
```
noudcorten, line 16: add :param: and :return: to comments

noudcorten (commented):
I really like the utils solution! Very nice to add 'simple' and globally used functions to an utils file.

utils.py
deborahvans: Is that necessary is there is no parameter and a generic result?

dema-trading-ai: @deborahvans yes it is necessary, because Codacy will block the PR if you do not

dema-trading-ai (approved):

utils.py
deborahvans: For the obvious cases Codacy allows one line comments (like I commited yesterday) As you can see Codacy is now blocking it again because the comment is too short to have multiple lines

noudcorten (approved):

PR #86, opened 2021-03-11
Fix open trades

### LINKED ISSUE #85: Open trades table wont show when there is no profit yet
When there is no profit yet the value is still None, which you cant round. Therefore, errors and table no-show

deborahvans:
# Fixes
Fixes #85

# What has been changed?
Value of None is changed to 0.0 before the table is generated

n/a

# Security Measures
n/a


backtesting/backtesting.py
```diff
@@ -136,6 +136,9 @@ def generate_open_trades_results(self, open_trades: [Trade]) -> typing.List[Open
         for trade in open_trades:
             if trade.max_drawdown is None:
                 trade.max_drawdown = 0.0
+            if trade.profit_percentage is None:
+                trade.profit_percentage = 0.0
+                trade.profit_dollar = 0.0
```
noudcorten, line 141: Why not just change the initialisation in the Trade class?

noudcorten (commented):
Nice find!

backtesting/backtesting.py
dema-trading-ai: This is also a good suggestion. What do you think is better? @deborahvans @noudcorten

deborahvans: I think thats a good idea! @noudcorten @dema-trading-ai I assumed there was a particular reason for the initialisation being None, but if there isnt changing it to 0 could save a lot statements like these throughout the document

dema-trading-ai: Alright! Let's do it in initialization. Do you have time to fix this right now? @deborahvans

PR #116, opened 2021-03-25
Adding a check for missing ticks

### LINKED ISSUE #84: Check for missing ticks / empty tick data
We want a hard check on possible missing ticks in downloaded data. This means that we have to check if all downloaded pairs have data for every needed tick. It could be that f.e. BTC/USDT has 1 missing tick on 13.00, while the other downloaded pairs do actually have this tick. In that case, we want to be notified using console error output. 

If you have any questions, hit us up on Discord.

pvdklei:
# Fixes

Fixes #84

# What has been changed?

In `Datamodule` I added a method that checks for missing ticks. If more than zero prints a warning. I did not find missing data, either by looking at the data or by this method, so it does not show anything right now. Warning what exact dates/ticks are missing seems too much, maybe when after adding a `--verbose` option in the cli. 

# Examples

`[WARNING] Pair 'USDT/BITC' contains 4 missing ticks`

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

# Additional Info

### DIFF
--- a/data/datamodule.py
+++ b/data/datamodule.py
@@ -1,6 +1,5 @@
 # Libraries
 import numpy as np
-from datetime import datetime, timedelta
 from pandas import DataFrame
 import pandas as pd
 import ccxt
--- a/data/datamodule.py
+++ b/data/datamodule.py
@@ -361,9 +361,10 @@ class DataModule:
             diff = np.setdiff1d(daterange, index)
             n_missing = len(diff)
 
-            if n_nan > 0 or n_missing > 0:
-                print(f"[WARNING] Pair '{pair}' contains {n_nan} rows with NaN values\n"
-                      f"[WARNING] and {n_missing} missing ticks")
+            if n_nan > 0:
+                print(f"[WARNING] Pair '{pair}' contains {n_nan} rows with NaN values")
+            if n_missing > 0:
+                print(f"[WARNING] Pair '{pair}' is missing {n_missing} ticks (rows)")
 
             
 
--- a/data/datamodule.py
+++ b/data/datamodule.py
@@ -366,7 +366,7 @@ class DataModule:
             n_missing = len(diff)
 
             if n_nan > 0 or n_missing > 0:
-                print(f"[WARNING] Pair '{pair}' contains {n_nan} rows with Nan values\n"
+                print(f"[WARNING] Pair '{pair}' contains {n_nan} rows with NaN values\n"
                       f"[WARNING] and {n_missing} missing ticks")
 
             
--- a/data/datamodule.py
+++ b/data/datamodule.py
@@ -93,6 +93,8 @@ class DataModule:
                 df = self.read_data_from_datafile(pair)
             self.history_data[pair] = df
 
+        self.check_for_missing_ticks()
+
         if self.same_backtesting_period():
             self.backtesting_module.start_backtesting(self.history_data, self.backtesting_from, self.backtesting_to)
         else:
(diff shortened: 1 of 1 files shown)


keebrev:
@pvdklei did you encounter missing ticks while testing this for this issue?

pvdklei:
@keebrev like I said, no. At least no null/NaN values. It could be that the exchange library you use already filters those out, and that indices (dates) are missing though.

dema-trading-ai:
> @keebrev like I said, no. At least no null/NaN values. It could be that the exchange library you use already filters those out, and that indices (dates) are missing though.

 In case we configured a 5 minute timeframe, ALL ticks should be 5m apart from eachother. Could you add another check to validate if this is the case for every candle?

noudcorten:
@dema-trading-ai @pvdklei I think a simple check is to create a list containing all the ticks that the dataframe should have (look at datamodule.py -> download_data_for_pair). Then compare this list with the index of the dataframe to see if there are some ticks missing. Simply said do something like:
```python
required_ticks = np.arange(start_date, data_to, self.timeframe_calc)
missing_ticks = set(required_ticks) - set(df.index)
```
Then, if len(missing_ticks) > 0 you know there is a tick missing (and which one)

pvdklei:
@keebrev I added something like @noudcorten suggested, should be oke now!

noudcorten:
@pvdklei Looks nice! But maybe it's nicer to first find out which method binance uses to fill up it's empty candles and then build a check around that. It could be on of these scenario's:
A. Empty candles are returned as timestamps with NaN as Open, High, Low, Close and Volume.
B. Empty candles are not returned and are missing in the data
C. There are no empty candles

So find which is scenario is true and then build the check around that, instead of covering it all.

keebrev:
@noudcorten I agree with you, although, since we do not know IF there is missing data and if there is, HOW we can find it, I think a cover all is doing the job for now. As soon as we know how missing data could be traced, we can update this code.

noudcorten:
@pvdklei @dema-trading-ai This PR is outdated. Index of the dataframe is now just the timestamp in miliseconds as an integer. So we're not working with datetime as index anymore.

pvdklei:
@noudcorten ccxt is unclear about what they do with missing values, so best we can do is guess that they add null values or delete rows. If they replaced missing values with something else than 'not a number', for instance with the mean, there is no way to find out which those were. 

Btw I changed it so it fit the new timestamp system

noudcorten:
@pvdklei Yes I agree with you! Then it's indeed better to keep it this way.

data/datamodule.py
```diff
@@ -339,3 +342,30 @@ def remove_backtesting_file(self, pair: str) -> None:
...
+
+            # any NaN?
+            n_nan = data.isnull().any(axis="columns").sum()
+            
+            # missing dates?
+            index = data.index.to_numpy().astype(int)
+            diff = np.setdiff1d(daterange, index)
+            n_missing = len(diff)
+
+            if n_nan > 0 or n_missing > 0:
```
noudcorten, line 364: Maybe change this to two if-statements, so that when one of the two triggers the print statement doesn't say '0 ...'.

pvdklei: Very true

data/datamodule.py
```diff
@@ -339,3 +342,30 @@ def remove_backtesting_file(self, pair: str) -> None:
...
+                              self.backtesting_to,
+                              self.timeframe_calc)
+
+        for pair, data in self.history_data.items():
+
+            # any NaN?
+            n_nan = data.isnull().any(axis="columns").sum()
+            
+            # missing dates?
+            index = data.index.to_numpy().astype(int)
```
noudcorten, line 360: .astype(int) can be removed here

pvdklei: Thought so as well, but results in weird warning and longer running time (for some reason)

noudcorten: Aah okay. Keep it then :)

noudcorten (approved):
Looks great! Solve codacy problem, then it's ok!

pvdklei:
@noudcorten done

keebrev (approved):

PR #117, opened 2021-03-25
Added ability to dynamically change the stoploss in a strategy

### LINKED ISSUE #75: Dynamic stoploss for every trade
We have to have to ability to configure stoploss (SL) on a per trade basis. This allows for configuring stoplosses on the go, based on indicators for example. This should be an option for more advanced users. 
The config should have the standard 'static' stoploss, which gets overridden in case a user specifies it when opening a trade. 

Things to do:
- Create method in strategy base class, method should return stoploss (float) for every new trade
- Baseclass method always returns config default SL
- Create SL calculation method in strategy, so it's configurable for every trade that opens + for every strategy
- Call SL calculation method on opening trade. Store SL in trade model
- Add property to trade model for storing trade specific SL
- Adjust methods in trading module for working with trade specific SL when this is turned on

pvdklei:
# Fixes

Fixes #75

# What has been changed?

You now have the option to change the stoploss per every tick by overriding the `stoploss` method on `Strategy`. If not provided, use configured stoploss value. This stoploss is also saved in every trade. 

I also changed some of the variable names and docstrings in `Strategy`. For instance the candle and indicator dataframes were both named `dataframe: DataFrame`. Ambiguous, and the type annotation already stated that is a `DataFrame`. 

# Examples

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

# Additional Info

### DIFF
--- a/backtesting/strategy.py
+++ b/backtesting/strategy.py
@@ -1,6 +1,7 @@
 import abc
 from pandas import DataFrame, Series
 from models.trade import Trade
+from typing import Optional
 
 # ======================================================================
 # Strategy-class is responsible for populating indicators / signals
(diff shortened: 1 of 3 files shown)


keebrev:
Could you make the stoploss method static / abstract? So people can implement this in their own strategy.

pvdklei:
@keebrev they already can, making it an abstract method would obligate users to override it (throw an error if they do not). We want to make it optional right. Static would mean that you cannot take in `self` as a parameter.

dema-trading-ai:
@pvdklei ah alright, then it's perfectly fine.

dema-trading-ai (approved):

keebrev (approved):

marijn111:
@pvdklei Does this change the stoploss for all open trades or just for a specific one?

PR #119, opened 2021-03-26
Add CLI support and made a configuration specification

### LINKED ISSUE #69: Command line argument support
If we want to change a variable we currently have to edit the config.json, which is a bit devious. Therefore it is nice to add command line support to easily change the variables or select a different strategy.

Things to do:
* Make every argument in the config.json configurable using the command line.
* Add --help argument to see which variables can be changed

pvdklei:
# Fixes

Fixes #69

# What has been changed?

You can now change certain (stoploss, strategy and starting capital, but can be very easily changed) config parameters through the CLI. I also changed the configuration to be based on a specification so we can easily add new parameters: name, what type they should be, if they should be cli configurable, what the possible inputs are, minimum/maximum values. 

# Examples

`$docker run dema-engine:alpha main.py --help`

```
usage: main.py [-h] [-cap STARTING_CAPITAL] [-sl STOPLOSS] [-s STRATEGY_NAME]

Dema Trading Engine

optional arguments:
  -h, --help            show this help message and exit
  -cap STARTING_CAPITAL, --starting-capital STARTING_CAPITAL
  -sl STOPLOSS, --stoploss STOPLOSS
  -s STRATEGY_NAME, --strategy-name STRATEGY_NAME
```

`$docker run dema-engine:alpha main.py -cap -100`
`[CONFIG ERROR] starting-capital = -100.0 is under the minimum value 1.0`

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

# Additional Info


dema-trading-ai:
@pvdklei could you maybe fix conflicting files?

config/validations.py
```diff
@@ -1,47 +1,57 @@
...
+    
+    if not good:
+        config_error(f"You passed an invalid type to the '{spec['name']}' parameter",
+                     f"This type should be a(n) {t}, it is {type(param_value)}")
 
-    :param config: the json config dictionary
-    """
-    for param, default_value in DEFAULT_PARAMS.items():
-        if config.get(param) == None:
-            config[param] = default_value
```
dema-trading-ai, line 43: How did you replace this?

config/validations.py
```diff
@@ -1,47 +1,57 @@
...
+def validate(config: dict, spec: list):
+    validate_by_spec(config, spec)
     validate_single_currency_in_pairs(config)
 
+def validate_by_spec(config, config_spec):
+    for param_spec in config_spec:
+        assert_given_else_default(config, param_spec)
+        assert_type(config, param_spec)
+        assert_in_options(config, param_spec)
+        assert_min_max(config, param_spec)
```
dema-trading-ai, line 13: Ah I see, you moved it here. Perfect dude, really nice

dema-trading-ai (approved):

keebrev (approved):

pvdklei:
automatic tests give error on `pip install ...`. Works fine on my computer, but takes very long, maybe thats it. Should I just merge?

keebrev:
@pvdklei can you fix codacy issues?

dema-trading-ai (changes requested):
@pvdklei When I use '0.5' as SL value, I get an 'config error' which shouldn't be there. Please fix before this can be merged

pvdklei:
@dema-trading-ai the stoploss in `config.json` was set at -0.5, so thought a negative value was requested and put the maximum allowed value 0. Do we want it to be positive or negative? We could also allow both, but then we should convert it to the same value in the config validation. 

@keebrev the codacy issues are unused imports in a `__init__.py` file, but thats just for convenient importing (e.g, you don't have to do `from config.validations import validate` but `from config import validate`). I can resolve it if you'd like but I suggest we just ignore those warnings.

keebrev:
@pvdklei  Ah alright, do not really mind. But if it's possible we should allow both. Anyway; the error I got is nog concerning negative of positive number, the error says I cannot use a Float as SL..

pvdklei:
@keebrev I cant replicate your bug, but I think it might be because I removed all the quotes ("...") around the numbers in the config file. So `"stoploss": -0.5` instead of `"stoploss": "-0.5". That way it is automatically casted to the right type and seems far more logical to me. I also swapped out `"backtesting-till-now": "False"` for `"backtesting-till-now": false`.

dema-trading-ai:
@pvdklei I am sorry, I think I used a wrong image for trying... Seems to work perfectly...

config/specification.json
```diff
@@ -0,0 +1,88 @@
...
+  {
+    "name": "exchange",
+    "default": "binance",
+    "options": ["binance"],
+    "type": "string"
+  },
+  {
+    "name": "timeframe",
+    "default": "5m",
+    "options": ["1m", "5m", "1h"],
```
noudcorten, line 11: @pvdklei Why don't we support more options? e.g. 15m is used quite often as well right?

config/specification.json
```diff
@@ -0,0 +1,88 @@
...
+    "cli": {
+      "short": "s"
+    }
+  },
+  {
+    "name": "strategies-folder",
+    "type": "string",
+    "default": "strategies"
+  }
+]
```
noudcorten, line 88: @pvdklei Is it possible to add an extra option based on PR #146 ? Once this PR is merged, we need to have the option 'stoploss-type' included. It has to handle the options stated in the PR description.

@dema-trading-ai @keebrev Do we want to wait with merging this PR before #146 is merged? What do you think?

config/cli.py
```diff
@@ -0,0 +1,23 @@
+import argparse
```
noudcorten, line 1: @pvdklei Do we need to update requirements.txt to be able to handle argparse?

config/specification.json
pvdklei: I thought the exchange had limited options, but I now see that we handle this with a regex and any value is possible. So I'll remove the options for now. I guess I can also add a regex entry in the spec.

config/specification.json
pvdklei: It's as easy a adding

```
[
  ..., 
  {
    "name": "stoploss-type",
    "type": "string",
    "options": ["dynamic", "standard", "trailing"]
  },
  ...
]
```

to the spec, and then it will do some basic type and options validation. Think that it's best if this change is made in #146 itself.

config/cli.py
pvdklei: No! `argparse` is in the python std

noudcorten: Ok! Nice :)

config/specification.json
noudcorten: Yes agreed. I will change it.

noudcorten (approved):

pvdklei:
@dema-trading-ai There is still an outdated 'changes requested', that's solved right? Then we can finally merge :)

PR #131, opened 2021-04-01
[Feat] Backtest optimisation

### LINKED ISSUE #104: Change backtesting strategy
Our current implementation for backtesting takes a lot of time to process all the ticks. This time issue is the result of our backtesting engine being designed as if it is 'live trading'. I think we can change this strategy to actual backtesting since all data is already available at the beginning of the backtest. To change the current implementation to an actual backtest, the engine needs to populate indicators, buy and sell signals only once (using all data). 

To do:
* Change backtesting strategy from 'live trading' to actual backtesting
     * Populate the indicators only once
     * Populate buy signals only once
     * Populate sell signals only once (incl. ROI and stoploss)
* Evaluate and improve that way that the 'Trade' class is currently used

It's hard for me to estimate how much time this is going to take to implement, but what do you guys think of the idea?

### LINKED ISSUE #122: Implement RapidJson
Have a look at https://pypi.org/project/python-rapidjson/ and what investigate what this could bring us in terms of speed.

### LINKED ISSUE #123: [Bug] fix bug when downloading data
Have a look at the screenshot.
screenshot

noudcorten:
# Fixes

Fixes #104
Fixes #122
Fixes #123 

# What has been changed?

* Significantly increased backtesting time
* Changed fee calculation to the correct implementation (was changed incorrectly)
* Population of indicators and signals is only executed once
* Population is moved from tradingmodule.py to backtesting.py
* Changed dataframe lookups to dictionary lookups
* Functions in strategy class now only receives the whole dataframe (not: passed_ticks + current_tick)
* .json files are now stored and retrieved using fastjson instead of the standard json library
* Added structure to the imports in the beginning of each file
* Fixed incorrect data downloading bug when 'backtesting-till-now' is true
* Changed method for converting dataframe to json and vice versa
* Dataframe is NOT using datetime as index anymore (changed to int)

# Examples

* dataframe lookups **vs** dictionary lookups -> speed increase of 146% (2:43m -> 1s)
* fastjson **vs** json -> speed increase of 21% (19.67s -> 15.57s)

# Security Measures

* Engine is not live trading anymore which can result in some incorrect strategies when creating algorithms (e.g. using .mean() without .rolling())
* Update the documentation so algo devs know what **not** to do when creating algo's.
* The dataframe index is changed to int instead of the previous datetime index. This makes certain PR's outdated (#116)

# Additional Info

Would deploy to main branch as soon as possible, since it increases the backtesting time **SIGNIFICANTLY**

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -1,12 +1,15 @@
+# Libraries
 from datetime import datetime, timedelta
-from backtesting.results import MainResults, OpenTradeResult, CoinInsights, show_signature
-from utils import *
-from models.trade import Trade
-from config.currencies import get_currency_symbol
-from config.load_strategy import load_strategy_from_config
 import typing
 from tqdm import tqdm
 import numpy as np
+
+# Files
+from backtesting.results import MainResults, OpenTradeResult, CoinInsights, show_signature
+from utils import df_to_dict, calculate_worth_of_open_trades, default_empty_dict_dict
+from models.trade import Trade
+from config.currencies import get_currency_symbol
+from config.load_strategy import load_strategy_from_config
 
 # ======================================================================
 # BackTesting class is responsible for processing the ticks (ohlcv-data)
@@ -57,7 +60,6 @@ class BackTesting:
 
         for tick in tqdm(ticks, total=len(ticks), ncols=75, desc='[TEST] Backtesting'):
             for pair in pairs:
-                # Get dict with data for current pair and pass to trading_module
                 pair_dict = data_dict[pair][tick]
                 self.trading_module.tick(pair_dict)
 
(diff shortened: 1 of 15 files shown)


marijn111:
This means that absolute indexes produces unexpected results. If the full dataframe is loaded at once, the user cannot use e.g. `dataframe['close'][-1]` anymore, because that will use data from the 'future'.

We should think about the consequences that this refactoring produces, such that we can inform the user as good as possible.

keebrev:
@marijn111 dataframe['close'][-1] shouldn't be possible anyway. This is something any user should be warned about in the documentation. Imo, shouldn't be fixed in the code.

keebrev (approved):

PR #153, opened 2021-04-15
[Feat] Coin plots

### LINKED ISSUE #73: Plotting possibilities after running backtest
A nice addition would be some kind of plots to visualize the following things after/during backtesting:
- Budget over time
- Buy/Sell moments on price graph of certain coin pair (OHLCV)
- Average profit of trades over time
- ... <insert additions here>

What do you guys think?
What should be changed on current implementation to match requirements for plotting?

deborahvans:
# Fixes

Fixes #73 

# What has been changed?
generate_indicators now also takes the str of the pair
generate_indicators now contains a plot function (which can be found in utils)

# Examples
image

NA

# Security Measures
NA

Plotly needs to be imported

# Additional Info
Originally I put the plot function in the data module, but if we want to be able to add certain indicators to the plots I think its best to be in the strategy. Like this in the stagey users can easily choose one of their indicators to plot.


strategies/my_strategy.py
```diff
@@ -28,6 +31,9 @@ def generate_indicators(self, dataframe: DataFrame) -> DataFrame:
         dataframe['ema5'] = ta.EMA(dataframe, timeperiod=5)
         dataframe['ema21'] = ta.EMA(dataframe, timeperiod=21)
 
+        indicator_plot = 'ema5'
+        plot_dataframe(pair, dataframe, indicator_plot)
```
noudcorten, line 35: I would move this to backtesting.py so that generate_indicators() stays clean for the algo creates. This change will also remove the extra added argument. Because generate_indicators() adds extra columns with the used indicators, you can easily extract these in your function using:
```pyton
added_indicators = dataframe.columns.values
```

noudcorten (commented):
Looks very nice! I would maybe change/add the following:
* Change the location of your plot_dataframe() call to keep generate_indicators() clean. 
* Plot the dataframe after the backtest is finished instead of plotting it before starting the backtest.
* Add some more signals to the plot (e.g. when 'buy' is triggered and when 'sell' is triggered).
* Add 'plot' boolean to config.json, so that people can choose if they want to see a plot or not (e.g. "plot" = "True" or "Plot" = "False").

PR #158, opened 2021-04-18
[Feat] Proper startup time

### LINKED ISSUE #23: Proper startup time
Because some indicators are calculated based on x amount of passed candles (f.e. SMA), we need a startup time to be able to generate the indicators properly.

Things to be done:
- Make startup time
- Check for presence in config file
- During the first x candles, send no signals / generate no indicators

Addition: if there's any possibility to find the appropriate startup time automatically, this would be perfect!

noudcorten:
# Fixes

Fixes #23 

# What has been changed?

* Since our engine is not live trading anymore, the if statement in the strategy class (```if len(dataframe) > self.min_candles:```)
has become useless. This is useless because we now populate our indicators once based on the whole dataframe, which will always pass this if statement. This is removed.
* Since generate_indicators will use certain timeframes to populate indicators, it will create NaN values until it reached the configured timeframe. If these values are used in the .loc function for generating buy or sell signals, it will always fail. Therefore, incorrect buy signals will never be generated and the 'startup time' is automatically defined.

# Examples

N/A

N/A

# Security Measures

N/A

N/A

# Additional Info

N/A

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -122,6 +122,10 @@ class BackTesting:
         overall_profit = ((budget - self.starting_capital) /
                           self.starting_capital) * 100
         max_seen_drawdown = self.calculate_max_seen_drawdown()
+        drawdown_from = datetime.fromtimestamp(max_seen_drawdown['from'] / 1000) \
+            if max_seen_drawdown['from'] != 0 else '-'
+        drawdown_to = datetime.fromtimestamp(max_seen_drawdown['to'] / 1000) \
+            if max_seen_drawdown['to'] != 0 else '-'
 
         return MainResults(tested_from=datetime.fromtimestamp(self.backtesting_from / 1000),
                            tested_to=datetime.fromtimestamp(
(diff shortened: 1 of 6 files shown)


backtesting/backtesting.py
```diff
@@ -87,6 +87,7 @@ def populate_signals(self) -> dict:
             indicators = self.strategy.sell_signal(indicators)
             stoploss = self.strategy.stoploss(indicators)
             self.config['stoploss'] = stoploss if stoploss else float(self.config['stoploss'])
+            indicators.dropna(inplace=True, axis=0)
```
marijn111, line 90: This will just remove the entire row if there is a `Nan` present right? Its meant for the first x amount of candles in which the indicators are not valid yet? That makes sense, however, it will thus also remove complete candles if there are other `Nan` values after the startup time right? That will make debugging harder when developing your strategy, since the candles of interest are dropped and not visible anymore when you e.g. print the dataframe for inspection.

marijn111 (approved):

dema-trading-ai (approved):

Issue #156, opened 2021-04-16
Revise dashboard overview


dema-trading-ai:
We need to improve the dashboard, add more, proper, data to it. Things to do:
* Change text: 'max drawdown' in individual trade to 'worst trade'
* Add 'best trade' to coin overview
* Include max open trades (from config.json)
* Put start/end date on same line
* Put start/end/profit balance on same line
* Fee percentage in / below 'paid fee' line
* Include timestamp and % for best/worst trade
* 
![image](https://user-images.githubusercontent.com/45246218/115021967-fcf6b900-9ebc-11eb-9096-6cb1641eb85b.png)

PR #171, opened 2021-04-28
[Fix] Correct drawdown calculation

### LINKED ISSUE #61: Add marketchange (avg + bitcoin)
We want to know the average marketchange for whitelisted coins in set backtesting period. 

Things to do:
- Make calculations for average marketchange for all whitelisted coins (printed %)
- Make calculations for average marketchange for bitcoin (as benchmark) (printed %)
- Include results in backtesting overview

### LINKED ISSUE #170: [Bug] - Avg trade duration
Encountered the following bug. Take a look at the backtesting period and the average trade duration for BTC.

Max open trade was 1.
Coinlist was BTC/USDT.
Timeframe was 1h.

Should be investigated and fixed.

image

noudcorten:
# Results of merging this

Fixes #170 
Fixes #61 

# What has been changed?

* Fixed wrong avg trade duration
* Changed the way 'spend_amount' (trade value for new trade) is defined. This will prevent the portfolio from becomming unbalanced (e.g. 1/4 budget in one trade and 3/4 budget in another trade)
    * Old: (1 / available_spaces) * budget
    * New: (1 / amount_of_pairs) * budget
* Fixed incorrect drawdown calculations for 'backtest results'
* Fixed incorrect drawdown calculations for 'per coin insights'
* Changed profit percentage columns for 'per coin insights'
    * Now consists of 'avg profit', 'cum profit' and 'total profit'
    * Profit percentages only include closed trades and not open trades (these are shown in 'left open trades' table)
* Renamed metrics in results table:
    * Max drawdown bad trade train -> most consecutive losses
    * Most drawdown 1 trade -> Worst trade
    * Seen drawdown from -> Max seen drawdown from
    * Seen drawdown to -> Max seen drawdown to
    * [Removed] Max drawdown trades
* Added average marketchange of BTC and whitelisted coins

# Examples

Schermafbeelding 2021-04-28 om 14 48 27

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

N/A

* The backtest results can differ with real life trading results, because of the refactured way of defining the 'spend_amount'.

* The way that realised drawdown is calculated on portfolio level and coin level.
* Intuition behind the current backtest results.

# Additional Info

I have added some features of #156, so that issue is almost finished.

### DIFF
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -40,7 +40,7 @@ class BackTesting:
         self.strategy = load_strategy_from_config(config)
 
     # This method is called by DataModule when all data is gathered from chosen exchange
-    def start_backtesting(self, data: dict, backtesting_from: int, backtesting_to: int) -> None:
+    def start_backtesting(self, data: dict, backtesting_from: int, backtesting_to: int, btc_marketchange_ratio: float) -> None:
         """
         Method formats received data.
         Method calls tradingmodule for each tick/candle (OHLCV).
@@ -51,12 +51,15 @@ class BackTesting:
         :type backtesting_from: int
         :param backtesting_to: 8601 timestamp
         :type backtesting_to: int
+        :param btc_marketchange_ratio: marketchange of BTC
+        :type btc_marketchange_ratio: float
         :return: None
         :rtype: None
         """
         self.data = data
         self.backtesting_from = backtesting_from
         self.backtesting_to = backtesting_to
+        self.btc_marketchange_ratio = btc_marketchange_ratio
         print('[INFO] Starting backtest...')
 
         data_dict = self.populate_signals()
@@ -177,7 +180,8 @@ class BackTesting:
                             tested_to=datetime.fromtimestamp(
                                self.backtesting_to / 1000),
                             max_open_trades=self.config['max-open-trades'],
-                            market_change=(market_change['all'] - 1) * 100,
+                            market_change_coins=(market_change['all'] - 1) * 100,
+                            market_change_btc=(self.btc_marketchange_ratio - 1) * 100,
                             starting_capital=self.starting_capital,
                             end_capital=budget,
                             overall_profit_percentage=overall_profit,
--- a/backtesting/backtesting.py
+++ b/backtesting/backtesting.py
@@ -72,7 +72,8 @@ class BackTesting:
         open_trades = self.trading_module.open_trades
         closed_trades = self.trading_module.closed_trades
         budget = self.trading_module.budget
-        self.generate_backtesting_result(open_trades, closed_trades, budget)
+        market_change = self.get_market_change(ticks, pairs, data_dict)
+        self.generate_backtesting_result(open_trades, closed_trades, budget, market_change)
 
     def populate_signals(self) -> dict:
         """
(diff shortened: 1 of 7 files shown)


backtesting/backtesting.py
```diff
@@ -256,50 +310,121 @@ def calculate_max_seen_drawdown(self) -> dict:
...
+                temp_seen_drawdown['peak'] = total_value
+                temp_seen_drawdown['bottom'] = total_value
+                temp_seen_drawdown['drawdown'] = 1.0  # ratio /w respect to peak
+                temp_seen_drawdown['from'] = tick
+                temp_seen_drawdown['to'] = tick
+            # Check if drawdown reached new bottom
+            elif total_value < temp_seen_drawdown['bottom']:
+                temp_seen_drawdown['bottom'] = total_value
+                temp_seen_drawdown['drawdown'] = temp_seen_drawdown['bottom'] / temp_seen_drawdown['peak']
+                temp_seen_drawdown['to'] = tick
```
keebrev, line 340: Shouldn't ``temp_seen_drawdown['to']`` be updated every tick (if there's drawdown?) Otherwise, you will only have a range from 1 tick when the max_drawdown of a certain drawdown period falls in the first tick. Right? While it could actually last for quite some ticks..? 

The tick with the most 'drawdown' isn't mostlikely the end of the period..?

keebrev:
@noudcorten If you are going to adjust something anyway. Is there a possibility to also add BTC marketchange as an extra overview stat? This could be useful. Have a look at #61

backtesting/backtesting.py
noudcorten: Yes this is a good comment. There can be two interpretations on how this backtesting period is defined:
1. (my interpretation) The backtesting period 'from' untill 'to' is defined as the start of a drawdown period untill the tick when max seen drawdown is reached. Following this intepretation the ['to'] does not need to be updated every tick, but is only updated when a new max seen drawdown is reached.
2. The backtesting period 'from' until 'to' is defined as the start of a drawdown period in which the max seen drawdown is included until the end of this period. In this interpretation the ['to'] indeed needs to be updated every tick and the final 'from' until 'to' then covers the complete drawdown period.

Which of these two interpretations is correct?

noudcorten (commented):
Responded to the remarks @keebrev

noudcorten:
> @noudcorten If you are going to adjust something anyway. Is there a possibility to also add BTC marketchange as an extra overview stat? This could be useful. Have a look at #61

@keebrev Yes this can be done. Do you want this in the same PR?

models/trade.py
```diff
@@ -15,31 +15,20 @@
...
         self.pair = ohlcv['pair']
         self.open = ohlcv['close']
-        self.current = ohlcv['close']
-        self.currency_amount = (trade_amount / ohlcv['close'])
         self.opened_at = date
+
+        self.fee = fee
+        self.starting_amount = spend_amount
+        self.lowest_seen_price = spend_amount
+        self.capital = spend_amount - (spend_amount * fee)  # apply fee
```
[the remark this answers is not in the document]
keebrev: Ah I see. Thx for the explanation.

PR #182, opened 2021-05-01
Fix a bug in the results when no trades are made

### LINKED ISSUE #181: Backtesting errors when no trades are made
image

WillemGerritzen:
# Results of merging this

Fixes #181

# What has been changed?

Added a check when the number of trades made is required that returns 1 if the variable == 0. On each instance of the usage, the variable is the denominator in a division.

# Examples

N/A

N/A

# Security Measures

This is exception handling in and of itself.

Might need to keep in mind if we change the way results are displayed. Otherwise nothing critical

N/A

# Additional Info


noudcorten (commented):
Good find! If no trades are made, the current placeholder value is now '1'. Maybe change this to '0' for cum_profit_percentage and to '-' for total_duration for a more logical backtest result.

sheridavandenbent (changes requested):
I agree with @noudcorten! Should be changed to '0' and '-' respectively. Otherwise, great catch!

WillemGerritzen:
Done! Thanks for the feedback :)

backtesting/results.py
```diff
@@ -107,13 +107,13 @@ def show(instances: typing.List['CoinInsights'], currency_symbol: str):
             stats.append([c.pair,
                         c.n_trades, 
                         round(c.market_change, 2),
-                        round(c.cum_profit_percentage / c.n_trades, 2),
+                        round(c.cum_profit_percentage / (c.n_trades if c.n_trades > 0 else 1), 2),
```
noudcorten, line 110: As mentioned before, I would change the 'else 1' to 'else 0' here.

keebrev: @WillemGerritzen could you change this?

geordle: Hi Guys, just joining in. I just wanted to draw your attention to this cool github feature.  
```suggestion
                        round((c.cum_profit_percentage / c.n_trades) if c.n_trades > 0 else 0, 2),
```

sheridavandenbent (approved):

keebrev (approved):

PR #192, opened 2021-05-04
Fix: Make sure the make command also runs with volume

sheridavandenbent:
# Results of merging this
Fixes issues that might arise from not running the engine with the docker volume. Running without the volume was useful in the past, but is currently unnecessary.

'make' command now has exactly the same result as the 'make runv' command.

# What has been changed?
in makefile, the runv command has been copied to the run command. runv command was left on purpose, to maintain backwards compatibility.

# Examples
N/A

N/A

# Security Measures
N/A

In the future the command 'runv' might be phased out because it is redundant.

Assumption: running without volume has no current uses.

# Additional Info
N/A

### DIFF
--- a/makefile
+++ b/makefile
@@ -7,7 +7,7 @@ build:
 	docker build . -t dema-engine:alpha
 
 run:
-	docker run --rm dema-engine:alpha
+	docker run --rm -v "$(shell pwd):/usr/src/engine/" dema-engine:alpha
 
 runv:
 	docker run --rm -v "$(shell pwd):/usr/src/engine/" dema-engine:alpha


marijn111 (approved):
Looks good. Nice idea about the backwards compatibility thing. We should schedule to remove one of the two in a next release though, as it is also confusing to have 2 commands doing the samen thing

sheridavandenbent:
Agree with removing runv in a while. We should throw a warning in there for the next release ("DEPRECATED: this command will be removed in the near future. Use 'make' instead of 'make runv'.") and then remove it in the release after that.

marijn111:
@sheridavandenbent Could you make an issue for that?

geordle (approved):

keebrev (approved):

PR #241, opened 2021-05-31
Fix/seen drawdown tests

sheridavandenbent:
# Results of merging this

Reintroduces seen drawdown into results overview and added thorough tests

# What has been changed?
Drawdown tests were moved to separate file, and seen drawdown tests were added

# Examples

# Security Measures

Explanation about seen drawdown and realised drawdown should be added to the wiki

# Additional Info


modules/stats/stats.py
```diff
@@ -36,6 +36,22 @@ def calculate_best_worst_trade(closed_trades):
...
+    total_value = 0
+    try:
+        total_value += value[tick]
+    except KeyError:
+        pass
+    try:
+        total_value += budget[tick]
+    except KeyError:
+        pass
+    return total_value
```
geordle, line 52: The purpose of the variable name `value` is a bit hard to grasp in this context. Maybe `open_order_value`?

I personally like this a bit better. Noticed that you just moved it BTW
```suggestion
def get_total_value_at_tick(value: dict, budget: dict, tick: int) -> float:
    return value.get(tick, 0) + budget.get(tick, 0)
```

geordle (changes requested):

modules/stats/stats.py
sheridavandenbent: Done in latest commit!

geordle (approved):

PR #249, opened 2021-06-01
Add timeperiod CLI support

### LINKED ISSUE #245: Add Timeperiod support to CLI
For allowing better programmatic use of the Engine, specifying the timeperiod via CLI should be possible

Flag could be `--timeperiod` for example.

Format could be 20200101-20200201

marijn111:
# Results of merging this

Fixes #245 

# What has been changed?

two flags have been added to the CLI arguments: `-from` and `-to`, corresponding respectively to `backtesting-from` and `backtesting-to`.

# Examples

# Security Measures

N/A

N/A

Changed the type from datetime to string, since the value entered is a string (right?). Only later gets the string transformed into a datetime object. 

Furthermore, no validations have been written, since the CLI basically overwrites the config.json. Afterwards, the config.json gets validated either way. 

# Packages added

N/A

# Additional Info


geordle (approved):

modules/setup/config/specification.json
```diff
@@ -26,11 +26,17 @@
   },
   {
     "name": "backtesting-from",
-    "type": "datetime"
+    "type": "string",
+    "cli": {
+      "short": "start"
```
geordle, line 31: IMO `start` => `from`  and `end` => `to` would be more in line with preexisting 
vars `backtesting-from` and `backtesting-to`. Might build on existing knowledge. Good enough though

geordle (approved):

modules/setup/config/specification.json
geordle: Resolve as you see fit

keebrev: @geordle would agree on this one.

marijn111:
@geordle I too agree that thats better, thanks. Modified accordingly

geordle (approved):

sheridavandenbent (approved):

keebrev:
@marijn111 I've updated this branche with `development`. PR can be merged now :)

PR #250, opened 2021-06-01
Remove graph zoom out due to buy signals with value zero

sheridavandenbent:
# Results of merging this

Fixes a bug that zooms the main plot out due to zero values in the buy & sell signals, instead of nan values

# What has been changed?

# Examples

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

# Packages added

# Additional Info


marijn111 (approved):
Since it doesn't modify the original dataframe inplace, appears to be a good solution. Plots look a lot better now. Nice!

PR #262, opened 2021-06-10
Feat/output

WesleyvanDooren:
# Results of merging this
Whenever you run the bot, the logs.json file will show you the results of each individual trade.

# What has been changed?
Only a new file has been added.

# Examples
   {
        "capital": 345.18791053243086,
        "close": 60427.55,
        "close_fee_amount": 0.8651326078507039
    }


config.json
```diff
@@ -18,7 +18,7 @@
   "fee": 0.25,
   "strategy-name": "MyStrategy",
   "strategies-folder": "strategies",
-  "plots": true,
+  "plots": false,
```
sheridavandenbent, line 21:
```suggestion
  "plots": true,
```

Don't push the changes to your config file!

trade_logs.json
```diff
@@ -0,0 +1,113 @@
+[
```
sheridavandenbent, line 1: Don't push this file to git! Add it to .gitignore - also it would be nice to add it to the data folder instead of in the root, just like the plots

PR #273, opened 2021-06-15
Fix/coin insight tests

sheridavandenbent:
# Results of merging this

Creates extra tests for coin insights and reintroduces the statistics in the result overview in two tables for easy overview.

# What has been changed?

# Examples

# Security Measures

# Packages added

# Additional Info

### DIFF
--- a/modules/output/results.py
+++ b/modules/output/results.py
@@ -115,13 +115,7 @@ class CoinInsights:
                         round(c.avg_profit_percentage, 2),
                         round(c.cum_profit_percentage, 2),
                         round(c.total_profit_percentage, 2),
-                        round(c.profit, 2),
-                        round(c.max_seen_drawdown, 2),
-                        round(c.max_realised_drawdown, 2),
-                        c.avg_trade_duration,
-                        c.roi, 
-                        c.stoploss,
-                        c.sell_signal])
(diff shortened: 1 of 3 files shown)


modules/output/results.py
```diff
@@ -115,13 +115,7 @@ def show(instances: typing.List['CoinInsights'], currency_symbol: str):
                         round(c.avg_profit_percentage, 2),
                         round(c.cum_profit_percentage, 2),
                         round(c.total_profit_percentage, 2),
-                        round(c.profit, 2),
-                        round(c.max_seen_drawdown, 2),
-                        round(c.max_realised_drawdown, 2),
-                        c.avg_trade_duration,
-                        c.roi, 
-                        c.stoploss,
-                        c.sell_signal])
```
keebrev, line 124: Did you remove this from the coininsights?

keebrev (approved):
Looks good to me. I like that you've moved the test-data / fixture outside the testmethods. Makes it lot better to understand / read.

modules/output/results.py
sheridavandenbent: Not really - they've been relocated to a table directly underneath so they won't be too large for the screen and mess up the formatting

PR #278, opened 2021-06-15
Made downloading of ohlcv concurrent

### LINKED ISSUE #272: Research possibility of multithreading/multiprocessing data download
Downloading coin data is not a dynamic task (right?), so therefore multithreading/multiprocessing should be possible to greatly speed up the process.

Investigate how much work this would take and what the benefits would be.
As reference: Downloading 5m data for 2 years for 12 coins can take 30 minutes..

geordle:
# Results of merging this
Fixes #272 

# What has been changed?
Made fetching of ohlcv concurrent

### DIFF
--- a/modules/setup/config/__init__.py
+++ b/modules/setup/config/__init__.py
@@ -4,6 +4,7 @@
 import json
 import re
 import sys
+from contextlib import asynccontextmanager
 from datetime import datetime
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


modules/stats/tradingmodule.py
```diff
@@ -157,13 +157,13 @@ def update_open_trades_value_per_timestamp(self, trade: Trade, ohlcv: dict) -> N
         Method is used to be able to track the open trades capitals per timestamp.
         It tracks the max seen point and the lowest seen point over all open trades.
         """
-        trade_opened_at = datetime.timestamp(trade.opened_at) * 1000
+        trade_opened_at = trade.opened_at.timestamp() * 1000
```
marijn111, line 160: This is for the consistency I assume?

geordle: `datetime.timestamp` was calling instance method as if it wasn't. This is kinda a no-no

modules/setup/config/__init__.py
```diff
@@ -26,47 +27,56 @@ class ConfigModule(object):
     timeframe: str
     timeframe_ms: int
 
-    def __init__(self, args):
+    @staticmethod
+    async def create(args):
+        config_module = ConfigModule()
```
keebrev, line 32: This is like a factory method now?

modules/setup/config/__init__.py
```diff
@@ -26,47 +27,56 @@ class ConfigModule(object):
...
+        return config_module
+
+    async def load_btc_marketchange(self):
         print("[INFO] Fetching marketchange of BTC/USDT...")
-        begin_data = self.exchange.fetch_ohlcv(symbol='BTC/USDT', timeframe=self.timeframe, since=self.backtesting_from, limit=1)
+        begin_data = await self.exchange.fetch_ohlcv(symbol='BTC/USDT', timeframe=self.timeframe,
+                                                     since=self.backtesting_from, limit=1)
         end_timestamp = int(np.floor(self.backtesting_to / self.timeframe_ms) * self.timeframe_ms) - self.timeframe_ms
-        end_data = self.exchange.fetch_ohlcv(symbol='BTC/USDT', timeframe=self.timeframe, since=end_timestamp, limit=1)
+        end_data = await self.exchange.fetch_ohlcv(symbol='BTC/USDT', timeframe=self.timeframe, since=end_timestamp,
```
keebrev, line 70: fetch_ohlcv is a method from `ccxt` right? Didn't know they supported async calls.

modules/setup/config/__init__.py
```diff
@@ -125,3 +135,14 @@ def config_from_to(exchange, backtesting_from: int, backtesting_to: int, backtes
...
+
+@asynccontextmanager
+async def create_config_module(args):
+    config_module = None
+    try:
+        config_module = await ConfigModule.create(args)
+        yield config_module
+    finally:
+        if config_module is not None:
+            await config_module.close()
```
keebrev, line 148: Could you briefly explain what this is for?

modules/setup/config/cctx_adapter.py
```diff
@@ -1,4 +1,4 @@
-import ccxt
+import ccxt.async_support as ccxt
```
keebrev, line 1: Cool! Didn't know.

modules/setup/config/__init__.py
keebrev: > nvm, found the answer already. Very cool.

geordle: yup, __init__ doesn;t allow `async`

modules/setup/config/__init__.py
geordle: Yup, __init__ doesn't allow `async`

modules/setup/datamodule.py
```diff
@@ -27,62 +28,66 @@
...
-                                               since=int(start_date),
-                                               limit=int(asked_ticks))
-
-            # Save timestamps and ohlcv info
-            index += [candle[0] for candle in result]  # timestamps
-            ohlcv_data += result
+            slice_request_payloads.append([asked_ticks, start_date])
             start_date += np.around(asked_ticks * self.config.timeframe_ms)
 
+        results = await asyncio.gather(*[self.exchange.fetch_ohlcv(symbol=pair,
```
keebrev, line 82: `asyncio.gather` is like `asyncio.wait`?

keebrev (approved):
I like it. `Zo is natoer...`

modules/setup/config/__init__.py
keebrev: Ah that's the reason. Clear, thanks.

modules/setup/config/__init__.py
geordle: Contextmanger, allows for that `with` syntax, similar to [using](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement) in c#. This is set up to dispose some cctx resources without interfering with normal flow & helping people remember to dispose

``` python

    async def run(self, args) -&gt; None:
          async with create_config_module(args) as config:
                setup_module = SetupModule(config)
```

geordle: [https://docs.python.org/3/library/contextlib.html](https://docs.python.org/3/library/contextlib.html)

modules/setup/datamodule.py
geordle: `wait` is a lower level primitive, more like some JS `Promise.{any, allSettled, all}` combined. `asyncio.gather(` is like `Promise.all`

keebrev: Clear. Thanks! 😄

PR #283, opened 2021-06-21
Remove show plots

geordle:
When developing these plots keep popping up, it's a bit annoying. They don't get opened when running in docker anyway. Would it be cool to remove this line?


sheridavandenbent (approved):
Sweet and simple, I approve! The plots are accessible in the data folder anyway, and most of our users use docker.

PR #280, opened 2021-06-17
Fix/max seen drawdown simple

sheridavandenbent:
# Results of merging this

simplifies the max seen drawdown for: portfolio, coin insight, and open trades.

PR #286, opened 2021-06-22
Fix/max realised drawdown simple

sheridavandenbent:
# Results of merging this

Simplifies the max realised drawdown, and the nr of losing trades and consecutive losses. Added tests for all three.

# What has been changed?

# Examples

- [x] PR depends on #280 (simple seen drawdown)

# Security Measures

# Packages added

# Additional Info

### DIFF
--- a/modules/stats/drawdown/for_portfolio.py
+++ b/modules/stats/drawdown/for_portfolio.py
@@ -1,5 +1,7 @@
 import pandas as pd
 
+from modules.stats.drawdown.drawdown import get_max_drawdown_ratio
+
 
 def get_max_seen_drawdown_for_portfolio(capital_per_timestamp: dict):
     max_seen_drawdown = {}
@@ -23,8 +25,6 @@ def get_max_seen_drawdown_for_portfolio(capital_per_timestamp: dict):
 
 def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict):
     df = pd.DataFrame.from_dict(realised_profits_per_timestamp, columns=['value'], orient='index')
-    df["drawdown"] = df["value"] / df["value"].cummax()
-
-    max_realised_drawdown = df["drawdown"].min()
+    max_realised_drawdown = get_max_drawdown_ratio(df)
 
     return max_realised_drawdown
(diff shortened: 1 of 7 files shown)


geordle (changes requested):

modules/stats/drawdown/per_coin.py
```diff
@@ -49,3 +49,32 @@ def add_trade_fee(df, fee_percentage, trades_closed_opened):
     for opened, closed in trades_closed_opened:
         df.loc[opened, "profit_ratio"] *= fee_ratio
         df.loc[closed, "profit_ratio"] *= fee_ratio
+
+
+def find_trade_timestamps(closed_pair_trades):
```
DemianVoorhagen6, line 54:
```suggestion
def get_trade_timestamps(closed_pair_trades):
```
get sounds more appropriate than find (as they are located already), I couldn't find what 'closed_pair_trades' entails exactly but 'pair' might not be needed? E.g. closed_trades would be sufficient?

DemianVoorhagen6 (commented):
Nice :) No other comments beside this, previous comments & Geordi's.

modules/stats/drawdown/for_portfolio.py
```diff
@@ -19,3 +19,12 @@ def get_max_seen_drawdown_for_portfolio(capital_per_timestamp: dict):
...
     return max_seen_drawdown
+
+
+def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict):
+    df = pd.DataFrame.from_dict(realised_profits_per_timestamp, columns=['value'], orient='index')
+    df["drawdown"] = df["value"] / df["value"].cummax()
+
+    max_realised_drawdown = df["drawdown"].min()
+
+    return max_realised_drawdown
```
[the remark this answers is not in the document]
sheridavandenbent: Done, good catch, thanks!

modules/stats/drawdown/per_coin.py
sheridavandenbent: closed_pair_trades is used here because it's the list of closed trades, but only for one pair. In other parts of the code, closed_trades holds the list of all closed trades of all pairs combined, so I do feel that the name closed_pair_trades adds more information

geordle (approved):

PR #296, opened 2021-06-28
[Feat] terminal width warning

noudcorten:
# Results of merging this

Fixes #

# What has been changed?

* Added warning if terminal width is too small, because this will cause incorrect display of the results.
* Engine now doesn't exit if no pairs are defined in the config.
* Renamed some output names.
* Compatible with PyCharm

# Examples

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

# Packages added

# Additional Info


modules/output/__init__.py
```diff
@@ -22,6 +26,16 @@ def output(self, stats: TradingStats):
 
         show_trade_anomalies(stats)
 
+        # check for correct window width
+        try:
+            terminal_width = os.get_terminal_size().columns
```
geordle, line 31:
```suggestion
            terminal_width = shutil.get_terminal_size().columns
```

geordle: Please read this [https://docs.python.org/3/library/os.html#querying-the-size-of-a-terminal](https://docs.python.org/3/library/os.html#querying-the-size-of-a-terminal). will also change usage a bit.

geordle: .

geordle (changes requested):

modules/output/__init__.py
noudcorten: When applying this change, the error message pops up when running it in PyCharm. But, the results display just fine as can be seen here: 
Schermafbeelding 2021-06-29 om 12 23 59

geordle: Hmm, just leave it I guess, let's merge

geordle (dismissed):

modules/output/__init__.py
```diff
@@ -25,6 +26,17 @@ def output(self, stats: TradingStats):
...
+            if terminal_width < 108:  # minimal terminal width
+                print_error("Your terminal width is too small. Increase "
+                            "terminal width to display results correctly.")
+        except OSError:
+            pass
+
+        show_signature()
+
         print("[INFO] Logging trades to " + FONT_BOLD + "data/backtesting-data/trades_log.json" + FONT_RESET + "...")
```
sheridavandenbent, line 40:
```suggestion
        print_info("Logging trades to " + FONT_BOLD + "data/backtesting-data/trades_log.json" + FONT_RESET + "...")
```

modules/output/results.py
```diff
@@ -8,21 +8,18 @@
...
-    print("%s   DemaTrading.ai BACKTESTING ENGINE IS SUBJECTED TO THE GNU AGPL-3.0 License %s" %
-          (FONT_BOLD, FONT_RESET))
-    print("%s   Copyright © 2021 - DemaTrading.ai%s" %
-          (FONT_BOLD, FONT_RESET))
-    print("================================================================================")
+    print_standard(
+        "================================================================================\n"
+        "   DemaTrading.ai BACKTESTING ENGINE IS SUBJECTED TO THE GNU AGPL-3.0 License\n"
+        "   Copyright © 2021 - DemaTrading.ai\n"
+        "================================================================================")
```
sheridavandenbent, line 22: Boldness of text is removed here - why? Should probably be reincorporated

noudcorten: I think copyright doesn't need to be that largely present in the results. The way it is now I think is fine.

modules/stats/tradingmodule.py
```diff
@@ -1,23 +1,23 @@
...
 # © 2021 DemaTrading.ai
 # ======================================================================
-from modules.stats.trade import SellReason, Trade
-from modules.stats.tradingmodule_config import TradingModuleConfig
 
 
 class TradingModule:
 
     def __init__(self, config: TradingModuleConfig):
-        print("[INFO] Initializing trading-module...")
```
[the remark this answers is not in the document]
noudcorten: This print statement had inconsistent placing (due to tqdm loading bars). This in combination with the statement not adding that much useful information gave me the conclusion of removing it

sheridavandenbent (approved):

sheridavandenbent (approved):

PR #297, opened 2021-06-29
Feat/longest shortest trade duration

sheridavandenbent:
# Results of merging this

Creates shortest, average, longest trade length statistic, plus tests, for both portfolio and per pair.


config.json
```diff
@@ -4,7 +4,7 @@
   "max-open-trades": 3,
   "starting-capital": 1000,
   "backtesting-from": "2021-03-11",
-  "backtesting-to": "2021-03-15",
+  "backtesting-to": "2021-05-15",
```
sheridavandenbent, line 7:
```suggestion
  "backtesting-to": "2021-03-15",
```

modules/output/results.py
```diff
@@ -106,6 +111,11 @@ def show(self, currency_symbol: str):
         console.print(table_grid)
 
     def create_trade_info_table(self, justification) -> Table:
+        avg_trade_duration = pandas.Series(self.avg_trade_duration).dt.round('1s')[0].to_pytimedelta() \
+            if self.avg_trade_duration != timedelta(0) else '-'
```
geordle, line 115: something like this:
```suggestion
        avg_trade_duration = timedelta(seconds=round(self.avg_trade_duration.total_seconds())) \
            if self.avg_trade_duration != timedelta(0) else '-'
```

geordle: Or this? 
```suggestion
        avg_trade_duration = Timedelta(self.avg_trade_duration).round("1s").to_pytimedelta() \
            if self.avg_trade_duration != timedelta(0) else '-'
```

test/stats/test_main_results.py
```diff
@@ -208,7 +209,7 @@ def test_dividing_assets():
 
 
 def test_n_trades():
-    """Given 'trades where made', 
+    """Given 'trades where made',
```
geordle, line 212:
```suggestion
    """Given 'trades were made',
```
?

modules/output/results.py
```diff
@@ -210,6 +225,11 @@ def show(instances: typing.List['CoinInsights'], currency_symbol: str):
         coin_signal_table = CoinInsights.create_coin_signals_table(justification)
 
         for c in instances:
+            avg_trade_duration = pandas.Series(c.avg_trade_duration).dt.round('1s')[0].to_pytimedelta() \
```
geordle, line 228: These lines look similar to those above. Maybe move this to a seperate function?

geordle (changes requested):
Nice!, just some aesthetics stuff.

test/stats/test_main_results.py
sheridavandenbent: Ohhh I can't believe I made a typo like that. Thanks!

modules/output/results.py
sheridavandenbent: Perfect! Went for the first one because it's better readable. Thank you!

modules/output/results.py
sheridavandenbent: Done!

geordle (approved):

PR #301, opened 2021-06-29
Update build_executables_mac.yml

geordle:
# Results of merging this
Executable deploy pipeline for MacOS was failing because it was referencing the wrong file. This should fix that


sheridavandenbent (approved):
Quick fix!! Thanks!

PR #305, opened 2021-06-29
Added output workspace

geordle:
# Results of merging this
New `<dir>` argument to init command fails to write to correct directory. When running in docker it now resolves relative to `output`


keebrev (approved):
Beautiful.

PR #312, opened 2021-07-01
Fix missingticks

### LINKED ISSUE #271: Check if data available for coins
Whenever you try to backtest over a period of time (and thus download coin data over that period) in which a certain coin didn't exist yet, the user will receive an error message similar to this:

```
Traceback (most recent call last):
  File "/usr/src/engine/main.py", line 25, in <module>
    main()
  File "/usr/src/engine/main.py", line 8, in main
    execute_for_args({
  File "/usr/src/engine/cli/arg_parse.py", line 39, in execute_for_args
    args.func(args)
  File "/usr/src/engine/main.py", line 17, in run_engine
    controller.run()
  File "/usr/src/engine/main_controller.py", line 18, in run
    algo_module = self.setup_module.setup()
  File "/usr/src/engine/modules/setup/__init__.py", line 14, in setup
    ohlcv_pair_frames = DataModule(self.config).load_historical_data()
  File "/usr/src/engine/modules/setup/datamodule.py", line 46, in load_historical_data
    df = self.read_data_from_datafile(pair)
  File "/usr/src/engine/modules/setup/datamodule.py", line 159, in read_data_from_datafile
    begin_index = df.index.get_loc(self.config.backtesting_from)
  File "/usr/local/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3082, in get_loc
    raise KeyError(key) from err
KeyError: 1590969600000
```

Implement a way of noticing that a coin didn't exist yet at the given starting time of backtesting.
Desired functionality imo: A one-liner warning during data-downloading saying that no data is available at the given starting time; downloading from the moment that there is data available. 
Backtesting should continue and take the coin in considering once there is data.

sheridavandenbent:
# Results of merging this

Fixes #271 

# What has been changed?

When downloading new data, if a coin didn't exist at the starting time yet, the missing timepoints in the beginning will be supplemented by NaN values. 

The market change is calculated by the first and last valid value (So skipping NaN)

# Security Measures

Maybe we should consider supplementing all missing ticks. However, we can't do it with NaN, TA-lib can't handle that. We could fill it with previous values, but then you can possibly get weird EMA and such calculation when there are 20 missing ticks.


modules/setup/datamodule.py
```diff
@@ -98,7 +98,7 @@ async def get_pair_data(self, pair, timeframe):
     async def load_markets(self) -> None:
         await self.exchange.load_markets()
 
-    async def download_data_for_pair(self, pair: str, data_from: int, data_to: int, save: bool = True) -> DataFrame:
+    async def download_data_for_pair(self, pair: str, data_from: int, data_to: int, save: bool = True, start: bool = True) -> DataFrame:
```
geordle, line 101:
```suggestion
    async def download_data_for_pair(self, pair: str, data_from: int, data_to: int, save: bool = True) -> DataFrame:
```

geordle (approved):

Issue #191, opened 2021-05-04
Optimize installation. 


dema-trading-ai:
We need to make a way that installing and using the engine is as simple as possible. This could possibly be achieved by creating a good working docker-compose file and creating proper explanation in the wiki. 

What do you think is the best way? 

Things we do not want:
- Users forced to re-download the whole repository when the engine updated.
- Users having to perform multiple steps before being able to run the new engine version.
- OS-dependent installations, we actually want 1 easy way of installing every necessary thing. 
- Installing binaries.

Things we DO want:
- Easy installation
- Preferably docker usage
- Easy docker-compose
- Step by step easy installation guide

PR #313, opened 2021-07-01
Feat/executables + plots cmd override

geordle:
# Results of merging this
Works towards #191 

* Fixes issues with multiprocessing.
* Changes one-file executable to one-dir (this speeds up start-up)
* Adds ability to set a cmd-line flag to override `plots` setting. (DevEx thing)

### DIFF
--- a/modules/output/plots.py
+++ b/modules/output/plots.py
@@ -1,7 +1,6 @@
 # Libraries
 import os
 from datetime import datetime
-from functools import partial
 from multiprocessing import Process
 
 import numpy as np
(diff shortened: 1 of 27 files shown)


modules/output/plots.py
```diff
@@ -1,4 +1,5 @@
 # Libraries
+import multiprocessing
```
geordle, line 2:
```suggestion
```

marijn111 (approved):
Looks good! Thanks for the walkthrough

PR #314, opened 2021-07-02
Feat/multiproces plots

geordle:
# Results of merging this
Run creation of plots in parallel. Speeds up plot creation by (.50 ^ number of pairs )

### DIFF
--- a/modules/output/plots.py
+++ b/modules/output/plots.py
@@ -1,4 +1,5 @@
 # Libraries
+import multiprocessing
 import os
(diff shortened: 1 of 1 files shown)


modules/output/plots.py
```diff
@@ -1,6 +1,9 @@
 # Libraries
+import multiprocessing
```
dema-trading-ai, line 2:
```suggestion
```

dema-trading-ai (approved):

PR #328, opened 2021-07-05
Feat/drawdown metric per coin

### LINKED ISSUE #274: Add Buy&Hold Drawdown metric per coin for risk comparison
We now have the market change (in profit) as metric in the backtesting report. This makes it possible to compare the profit. However, it would also be nice to have a way of comparing the risk of the bot with the market changes. This basically boils down to calculating the DD for a buy&hold strategy and report that too (in the per coin insights)

sheridavandenbent:
# Results of merging this

Fixes #274

# What has been changed?
BTC, coin, and market seen drawdown is added in the statistics. When BTC is present in the pairs, the pair data is used; if BTC is not present in the pairs, BTC's data is downloaded separately.

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,7 +13,7 @@
     "60": 4,
     "120": 3
   },
-  "pairs": ["BTC/USDT", "LTC/USDT", "ETH/USDT"],
+  "pairs": ["LTC/USDT", "ETH/USDT"],
   "currency": "USDT",
   "fee": 0.25,
   "strategy-name": "MyStrategy",
--- a/config.json
+++ b/config.json
@@ -13,7 +13,7 @@
     "60": 4,
     "120": 3
   },
-  "pairs": ["LTC/USDT", "ETH/USDT"],
+  "pairs": ["BTC/USDT", "LTC/USDT", "ETH/USDT"],
   "currency": "USDT",
   "fee": 0.25,
   "strategy-name": "MyStrategy",
--- a/modules/setup/datamodule.py
+++ b/modules/setup/datamodule.py
@@ -13,6 +13,7 @@ from cli.print_utils import print_info, print_error, print_warning
 
 # Files
 from modules.setup.config import ConfigModule
+from modules.stats.drawdown.drawdown import get_max_drawdown_ratio
 from utils.utils import get_ohlcv_indicators
 
 # ======================================================================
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -13,7 +13,7 @@
     "60": 4,
     "120": 3
   },
-  "pairs": ["BTC/USDT", "LTC/USDT", "ETH/USDT"],
+  "pairs": ["LTC/USDT", "ETH/USDT"],
```
sheridavandenbent, line 16:
```suggestion
  "pairs": ["BTC/USDT", "LTC/USDT", "ETH/USDT"],
```

modules/stats/stats.py
```diff
@@ -66,6 +69,19 @@ def calculate_trade_durations(closed_trades):
...
+    market_drawdown = {}
+    total_drawdown = [0] * len(data_dict[pairs[0]])
+    for pair in pairs:
+        values_list = [data_dict[pair][d].get('close') for d in data_dict[pair]]
+        coin_values = DataFrame(values_list, columns=['value'])
+        market_drawdown[pair] = get_max_drawdown_ratio(coin_values)
+        values_list = [x / values_list[0] for x in values_list]
+        total_drawdown = map(lambda x, y: x + y, total_drawdown, values_list)
+    market_drawdown['all'] = get_max_drawdown_ratio(DataFrame(total_drawdown, columns=['value']))
+    return market_drawdown
```
geordle, line 82: This looks like it works! Performance wise this might get a bit slow when testing longer periods though (we'll cross that bridge when we get there, using `df` might be a bit quicker ). 

What do you think of the following names?,  
```suggestion
    market_drawdown = {}
    pair_profit_sum = [0] * len(data_dict[pairs[0]])
    for pair in pairs:
        close_prices = [data_dict[pair][d].get('close') for d in data_dict[pair]]
        close_prices_df = DataFrame(values_list, columns=['value'])
        market_drawdown[pair] = get_max_drawdown_ratio(close_prices_df)
        profit_ratios = [x / values_list[0] for x in values_list]
        pairs_profit_ratios_sum = map(lambda x, y: x + y, pair_profit_sum, profit_ratios)
    market_drawdown['all'] = get_max_drawdown_ratio(DataFrame(total_drawdown, columns=['value']))
    return market_drawdown
```

geordle (dismissed):

modules/setup/datamodule.py
```diff
@@ -51,6 +56,19 @@ async def load_btc_marketchange(self):
...
         return end_close_value / begin_close_value
 
+    async def load_btc_drawdown(self, df: dict):
+        print_info("Fetching market drawdown of BTC/USDT...")
+
+        if 'BTC/USDT' in df.keys():
+            bitcoin_df = df.get('BTC/USDT')
+        else:
+            bitcoin_data = await self.get_pair_data('BTC/USDT')
+            bitcoin_df = bitcoin_data[1]
```
geordle, line 66:
```suggestion
            pair, bitcoin_df = await self.get_pair_data('BTC/USDT')
```

sheridavandenbent: Thanks, didn't know why I didn't think of that!

geordle (dismissed):

modules/stats/stats.py
sheridavandenbent: Variables are renamed! Thanks for the suggestions!

geordle (approved):

PR #329, opened 2021-07-05
Fix small bug while downloading extra candles

deborahvans:
# Results of merging this

Fixes #
Bug when downloading extra candles

# What has been changed?

Convert all index to str again before locating them.
Atm old ones were int & new added str


geordle (approved):
Nice, good catch & good fix! Let's merge this.

Just one thing to note. While looking at this issue I noticed this flip/flopping between index types  is causing some confusion. I'd rather have the index be an integer everywhere to raise understandability a bit (should make lookups a tiny bit faster too). @deborahvans, after merging this, would you be willing to review a follow-up PR that removes the `str` indexes?

PR #330, opened 2021-07-05
Fix/change df index to int

geordle:
Follow-up on #329 

# What has been changed?
Changed df index from `str` to `int`. 

FYI: `numeric` indexed series `[]` acts like `.loc[]` not `.iloc[]`, hence the explicit `.iloc[]` in `plots.py`

- [ ] PR depends on #329


deborahvans (approved):
Nice! The str's indeed felt confusing and unnecessary to me as well.

Looking good to me!

PR #332, opened 2021-07-07
Additional pairs functionality

marijn111:
# Results of merging this

Fixes #

# What has been changed?

Functionality to have additional pairs in a strategy has been added. It makes use of the normal data module, and therefore downloads and caches the extra requested candles. Furthermore a function has been added that merges the additional data with the dataframe in such a way that a look-into-the-future bias is prevented. 

I tried to implement it in such a way that it requires little extra effort (such as importing extra stuff etc), and ensures backwards compatibility. So users that don't want/need additional data, can still use their current strategy files. 
# Examples

An example has been added to the advanced strategy example file. 

- [ ] PR depends on #
- [ ] Script to be run

# Security Measures

N/A

Due to performance reasons, it is not recommended to request multiple timeframe for the same coin, but rather resample such that the different timeframes are achieved, since otherwise each pair needs downloading etc.

A full example section needs to be added to the wiki that explains this. 

# Packages added

N/A

# Additional Info

N/A


docker-compose.yml
```diff
@@ -1,8 +1,9 @@
...
+    image: dematrading/engine:develop
     volumes:
-      - "./strategies:/usr/src/engine/strategies"
-      - "./config.json:/usr/src/engine/config.json"
-      - "./data/backtesting-data:/usr/src/engine/data/backtesting-data"
\ No newline at end of file
+#      - "./strategies:/usr/src/engine/strategies"
+#      - "./config.json:/usr/src/engine/config.json"
+#      - "./data/backtesting-data:/usr/src/engine/data/backtesting-data"
+        - "./:/usr/src/engine"
```
sheridavandenbent: Did you mean to push these changes too? I don't think it's a good idea to set the standard image to develop ;)

modules/algo/backtesting.py
```diff
@@ -25,12 +25,13 @@ class BackTesting:
...
     df = {}
 
-    def __init__(self, data: dict, config_module: ConfigModule, strategy: Strategy):
+    def __init__(self, data: dict, config_module: ConfigModule, strategy: Strategy, additional_data):
         self.config = config_module
         self.starting_capital = config_module.starting_capital
         self.currency_symbol = config_module.currency_symbol
         self.strategy = strategy
         self.data = data
+        self.additional_data = additional_data
```
sheridavandenbent, line 34:
```suggestion
        self.additional_pairs_data = additional_data
```

Maybe changing the variable names to this (or something like it) makes it easier to understand what kind of data this is

docker-compose.yml
marijn111: Woops thats some debug stuff that got left in, thanks for noticing!

backtesting/strategy.py
```diff
@@ -64,3 +66,33 @@ def stoploss(self, dataframe: DataFrame) -> DataFrame:
         :rtype: DataFrame
         """
         return
+
+    def join_additional_data(self, dataframe, additional, timeframe, timeframe_additional, ffill=True):
+        """
+        This function is responsible for joining the additional data to the original dataframe
+        """
```
sheridavandenbent, line 73:
```suggestion
```

I would leave out the comment as it's implied in the name of the function

resources/setup/strategies/my_strategy_advanced.py
```diff
@@ -12,13 +12,29 @@ class MyStrategyAdvanced(Strategy):
     This is an example custom strategy for advanced users, that inherits from the main Strategy class
     """
 
-    def generate_indicators(self, dataframe: DataFrame) -> DataFrame:
+    def additional_pairs(self):
+        """
+        Specify The additional pairs in tuple format; ("pair", "timeframe")
+        :return: list
+        """
```
sheridavandenbent, line 19:
```suggestion
    def additional_pairs(self) -> list:
        """
        Specify The additional pairs in tuple format; ("pair", "timeframe")
        """
```

sheridavandenbent: If you can put it in the function instead of in comments, that's the way to go!

backtesting/strategy.py
```diff
@@ -64,3 +66,33 @@ def stoploss(self, dataframe: DataFrame) -> DataFrame:
...
+        add_df = additional.copy()
+        pair = add_df['pair'].unique()[0]
+
+        timeframe_minutes = parse_timeframe(timeframe)
+        timeframe_minutes_additional = parse_timeframe(timeframe_additional)
```
noudcorten, line 78: Don't know if these variable names should contain 'minutes', since parse_timeframe() returns miliseconds.
```suggestion
        timeframe_ms = parse_timeframe(timeframe)
        timeframe_ms_additional = parse_timeframe(timeframe_additional)
```

backtesting/strategy.py
```diff
@@ -64,3 +66,33 @@ def stoploss(self, dataframe: DataFrame) -> DataFrame:
         :rtype: DataFrame
         """
         return
+
+    def join_additional_data(self, dataframe, additional, timeframe, timeframe_additional, ffill=True):
```
noudcorten, line 70:
```suggestion
    def join_additional_data(self, dataframe, additional, timeframe_configured, timeframe_additional, ffill=True):
```
Would make it more clear what 'timeframe' is here.

modules/setup/datamodule.py
```diff
@@ -68,17 +68,21 @@ async def load_btc_drawdown(self, df: dict):
...
-        dataframes = await asyncio.gather(*[self.get_pair_data(pair) for pair in self.config.pairs])
+    async def load_historical_data(self, pairs, check_period=True) -> dict:
+        dataframes = await asyncio.gather(*[self.get_pair_data(pair, self.config.timeframe) if not isinstance(pair, tuple)
+                                            else self.get_pair_data(pair[0], pair[1]) for pair in pairs])   # if tuple then additional pair and timeframe comes specified with it
 
         history_data = {key: value for [key, value] in dataframes}
 
         self.warn_if_missing_ticks(history_data)
-        if not is_same_backtesting_period(history_data):
+        if not is_same_backtesting_period(history_data) and check_period:
```
sheridavandenbent, line 78:
```suggestion
        if check_backtesting_period and not is_same_backtesting_period(history_data):
```
The first one will get checked first, so if it's false the code doesn't have to do the is_same_backtesting_period function! Also I'd rename check_period to check_backtesting_period, just to make it a tad clearer!

sheridavandenbent (changes requested):
Looks like it's working correctly, nice job!

Do we want to add an example to the MyStrategyAdvanced to show what you could use the additional pairs data for?

modules/setup/datamodule.py
marijn111: Good one, thanks!

marijn111:
@sheridavandenbent There is an example in the advanced strategy file

backtesting/strategy.py
```diff
@@ -64,3 +66,40 @@ def stoploss(self, dataframe: DataFrame) -> DataFrame:
...
+    def join_additional_data(self, dataframe, additional_pair, original_timeframe, timeframe_additional, ffill=True):
+        """
+        This function is responsible for joining the additional data to the original dataframe. This is only possible
+        if the additional timeframe is larger than the original timeframe. Furthermore some precautions have been taken
+        such that candles are matched together without a looking in the future bias.
+        :param dataframe: original Dataframe
+        :param additional_pair: Dataframe of the additional pair
+        :param original_timeframe: timeframe of original dataframe
+        :param timeframe_additional: timeframe of additional dataframe
+        :param ffill: Wether to forward fill the joined dataframe (recommended)
```
sheridavandenbent, line 79:
```suggestion
        :param ffill: Whether to forward fill the joined dataframe (recommended)
```

sheridavandenbent (approved):

PR #334, opened 2021-07-07
Added 3 features + dependency upgrade

rhulshof:
- Added support for multiple traces
- Added support for creating an Excel tearsheet file
- Added config option for maximum exposure (to be able to spend i.e. 100% of balance into 1 single coin)
- Upgrade TA-Lib dependency to version 0.4.20 (from 0.4.19)


modules/stats/tradingmodule.py
```diff
@@ -102,7 +103,7 @@ def open_trade(self, ohlcv: dict, data_dict: dict) -> None:
             return
 
         # Define spend amount based on realised profit
-        spend_amount = (1. / self.amount_of_pairs) * self.realised_profit
+        spend_amount = (self.max_exposure / 100.) * self.realised_profit
```
geordle, line 106: Unfortunately we'll need to think about this a bit more before we can add this feature. In its current state it would significantly impact existing strategies. My thought is that the previous implementation was the way it is so that strategies on pairs would always have enough balance to get into position (this makes results more reproducible. Just think about the order in which trades are executed having effect on results in this case). Could you maybe explain the behaviour you're trying to achieve here? Maybe we can come up with a different solution together that fits your needs (exposure per pair?, dynamic stake amount?) cc: @DemianVoorhagen6 @sheridavandenbent

modules/output/plots.py
```diff
@@ -17,8 +17,7 @@
 def plot_sizes(subplot_indicator, df):
     rows = 1
     for ind in subplot_indicator:
-        if ind in df.columns.values:
-            rows += 1
+        rows += 1
```
geordle, line 20: `rows = 1 + len(subplot_indicator)`

modules/output/__init__.py
```diff
@@ -79,3 +86,11 @@ def log_trades(stats: TradingStats):
...
     with open('./data/backtesting-data/trades_log.json', 'w', encoding='utf-8') as f:
         f.write(trades_json)
+
+def create_tearsheet(stats: TradingStats):
+    dict_count = len(stats.trades)
+    df = pd.DataFrame(stats.trades[0].__dict__, index=[0])
+    for i in range(1,dict_count-1):
+        df = df.append(stats.trades[i].__dict__, ignore_index=True)
+
+    df.to_excel('data/backtesting-data/tearsheet.xlsx')
```
geordle, line 96: I needed to install a dependency for this called `openpyxl`. Maybe this needs to be added to `requirements.txt`

geordle (commented):
Awesome! Code is good & even follows our "undocumented" guidelines, kudos. There are some things regarding max_exposure that will take some deliberation but rest looks good.

modules/stats/tradingmodule.py
rhulshof: The goal is to analyze what the correlation is between several coins. I.e. one can decide to spend 100% of the balance on whichever coin displayes a buy signal first. Or one can decide to spend 50% per coin, whilst watching at 3 coins separately. So whichever 2 coins give their buy signal first, will end up with 50% of the balance. The backtester already supports this by displaying a "cannot buy" message in the terminal message, whenever the balance is already used on other coins.

Despite the above, I totally agree with your objections. It indeed would break existing setups. What I can do, is make sure the behavior remains as it was whenever the configuration setting is missing and/or None. Only when it actually has a percentage value, it will follow the above logic. This way, backtesters are completely in control. Do you agree? If yes, I will update the pull request.

geordle: That `cannot buy` message should not be possible in recent versions (from 2 months ago onwards) of the engine.

I'm OK with that but, what if we changed the `max_exposure` implementation a little bit. Currently, when wanting to distribute balance across trading pairs you'd have to manually  `100 / amount_of_pairs` and fill that for the `max_exposure` variable. Everytime you change your pairs you'd have to update this variable. If we change this to `exposure_per_trade` (note the ~~max~~, since it will actually decide exposure, not cap it) we could spread funds evenly across pairs with a value of `1.`, flatten sways by lowering it `FI: .5` or achieve what you're trying to do by setting it by raising `2`.

the adjusted formula would be something in the lines of
```python
exposure_per_pair = config["exposure_per_trade"] or 1.
splits_amount = min(self.max_open_trades, self.amount_of_pairs) # recent change
spend_amount = ((1. / splits_amount) * exposure_per_pair) * self.realised_profit
```

What do you think?

geordle: I'm not yet decided on whether we should use ratio or percentages, or on the naming

modules/output/__init__.py
rhulshof: Indeed. Will do.

modules/stats/tradingmodule.py
rhulshof: Stoploss setting already uses percentages, not ratio's. Therefore used the same here. Should be the way to go right?

rhulshof: I like your change and will build it accordingly.

modules/output/plots.py
rhulshof: Clever :)

rhulshof:
Request was updated. It now includes all discussed topics.

modules/public/trading_stats.py
```diff
@@ -11,6 +11,7 @@ class MainResults:
     tested_from: datetime
     tested_to: datetime
     max_open_trades: int
+    exposure_per_trade: int
```
geordle, line 14:
```suggestion
    exposure_per_trade: float
```

modules/stats/stats_config.py
```diff
@@ -10,10 +10,12 @@ class StatsConfig:
     stoploss: float
     stoploss_type: str
     max_open_trades: int
+    exposure_per_trade: int
```
geordle, line 13:
```suggestion
    exposure_per_trade: float
```

modules/stats/tradingmodule_config.py
```diff
@@ -8,6 +8,7 @@
 class TradingModuleConfig:
     fee: float
     max_open_trades: int
+    exposure_per_trade: int
```
geordle, line 11:
```suggestion
    exposure_per_trade: float
```

geordle (approved):

config.json
```diff
@@ -2,6 +2,7 @@
   "exchange": "binance",
   "timeframe": "5m",
   "max-open-trades": 3,
+  "exposure-per-trade": 1,
```
geordle, line 5:
```suggestion
  "exposure-per-trade": 100,
```

config.example.json
```diff
@@ -2,6 +2,7 @@
   "exchange": "binance",
   "timeframe": "5m",
   "max-open-trades": "3",
+  "exposure-per-trade": "100",
```
geordle, line 5:
```suggestion
  "exposure-per-trade": 100,
```

modules/setup/config/__init__.py
```diff
@@ -72,7 +73,9 @@ async def create(args):
         config_module.stoploss = config["stoploss"]
         config_module.stoploss_type = config["stoploss-type"]
         config_module.max_open_trades = config["max-open-trades"]
+        config_module.exposure_per_trade = config["exposure-per-trade"]
```
geordle, line 76:
```suggestion
        config_module.exposure_per_trade = config["exposure-per-trade"] / 100.
```

test/stats/stats_test_utils.py
```diff
@@ -10,6 +10,7 @@
 StatsModuleFactory = [[], StatsModule]
 
 max_open_trades = 3
+exposure_per_trade = 100
```
geordle, line 13:
```suggestion
exposure_per_trade = 1.
```

modules/output/results.py
```diff
@@ -176,6 +175,7 @@ def create_settings_table(self: MainResults, currency_symbol, justification, tes
                            f"{round(self.starting_capital, 2)} {currency_symbol}")
     settings_table.add_row("Fee percentage", f"{self.fee} %")
     settings_table.add_row("Max. open trades", str(self.max_open_trades))
+    settings_table.add_row("Max. exposure per trade", str(self.exposure_per_trade))
```
geordle, line 178:
```suggestion
    settings_table.add_row("Exposure per trade", str(self.exposure_per_trade * 100) + " %")
```

geordle (changes requested):
There are a few things left regarding that exposure ratio format. think percentages would be nice for user input, and internally normalized (1.) ratios are nice because they"re normal :). i think tests should also succeed after these changes

test/stats/stats_test_utils.py
```diff
@@ -23,6 +24,7 @@ class StatsFixture:
     def __init__(self, pairs: list):
         self.stats_config = StatsConfig(
             max_open_trades=max_open_trades,
+            exposure_per_trade=exposure_per_trade,
```
geordle, line 27:
```suggestion
            exposure_per_trade=exposure_per_trade,
             tearsheet=False,
```

rhulshof:
On my Windows machine all pytest's are failing. Even when running it on the "main" branch. All other issues I have fixed. Do you know why it's still failing?

geordle:
> On my Windows machine all pytest's are failing. Even when running it on the "main" branch. All other issues I have fixed. Do you know why it's still failing?

Hmm, weird. In pipelines and local they seem to work for me. If end-to-end tests succeed I'm completely fine with the changes. Thanks a lot!

geordle:
> On my Windows machine all pytest's are failing. Even when running it on the "main" branch. All other issues I have fixed. Do you know why it's still failing?

I'll take a look on windows as soon as I can

modules/setup/config/__init__.py
```diff
@@ -72,7 +74,9 @@ async def create(args):
         config_module.stoploss = config["stoploss"]
         config_module.stoploss_type = config["stoploss-type"]
         config_module.max_open_trades = config["max-open-trades"]
+        config_module.exposure_per_trade = config["exposure-per-trade"] / 100.
         config_module.plots = config["plots"]
+        config_module.tearsheet = config["tearsheet"]
```
geordle, line 79: Ah yeah, this is a breaking change. existing configs will struggle on this (hence the failing end-2-end tests).
The following will access safely and provide a default value.
```suggestion
        config_module.exposure_per_trade = config.get("exposure-per-trade", 100.) / 100.
        config_module.plots = config["plots"]
        config_module.tearsheet = config.get("tearsheet", False)
```

geordle (changes requested):

modules/setup/config/__init__.py
geordle: I'm checking pytest on windows now but should stop us from merging

geordle:
Tests on windows are failing because timestamp is broken for one day past epoch on windows. https://stackoverflow.com/a/67904074

Might fix this later

geordle (approved):

PR #335, opened 2021-07-07
Winning weeks

### LINKED ISSUE #200: Add overview of winning days/weeks/months per pair
Currently we show the total profit per coin (E.G. LTC/USDT 100%, ETH/USDT -30%, etc). It would be nice if we could see something like: Winning weeks: LTC/USDT (40 won/12 lost), ETH/USDT (5 won/47 lost). This might be useful to show losing pairs. We have to discuss what period to do and which implementation, but this is the first concept idea.

noudcorten:
# Results of merging this

Fixes #200 

# What has been changed?

* Implemented winning weeks on portfolio and coin level
* Created loading bar when calculating statistics
* Restructured double code in drawdown/per_coin calculations (also to make it easier to use in winning weeks calculation)
* Moved calculations for marketchange, trades (best/worst/etc.) and winning weeks from the stats.py file to separate files in a new folder ./stats/metrics/


modules/output/results.py
```diff
@@ -293,6 +302,8 @@ def create_coin_metrics_table(justification) -> Table:
         coin_metrics_table.add_column("Max. seen drawdown (%)", justify=justification)
         coin_metrics_table.add_column("Max. realised drawdown (%)",
                                       justify=justification)
+        coin_metrics_table.add_column("Winning weeks (W/D/L)",
```
sheridavandenbent, line 305: Same comment as above

modules/stats/drawdown/per_coin.py
```diff
@@ -1,16 +1,21 @@
 import pandas as pd
 import numpy as np
 
-from modules.stats.drawdown.drawdown import get_max_drawdown_ratio
 from modules.stats.trade import Trade
 
 
-def get_max_seen_drawdown_per_coin(signal_dict, closed_pair_trades: [Trade], fee_percentage: float):
+def get_cum_profit_ratio_per_coin(signal_dict, closed_pair_trades: [Trade], fee_percentage: float,
```
sheridavandenbent, line 7: If you don't calculate the drawdown here anymore, it should not be in the drawdown folder still

modules/stats/drawdown/per_coin.py
```diff
@@ -1,16 +1,21 @@
...
     trades_opened_closed_timestamps = map_trades_to_opened_closed_timestamps(closed_pair_trades)
 
     values = signal_dict.values()
     df = pd.DataFrame(values).set_index("time")
 
+    # Check if we want to calculate realised drawdown
+    if realised_drawdown:
+        trade_timestamps = get_trade_timestamps(closed_pair_trades)
+        df = pd.concat([df, trade_timestamps], axis=1, join="inner")
+
```
sheridavandenbent, line 18: I don't agree with mixing the function here - sure, it's dryer, but it's also less readable. From the name of the function it's not clear what you're calculating exactly, best to split it back up in two more methods imho

modules/stats/stats.py
```diff
@@ -110,7 +52,8 @@ def generate_backtesting_result(self, market_change: dict, market_drawdown: dict
 
         trading_module = self.trading_module
 
-        coin_results = self.generate_coin_results(trading_module.closed_trades, market_change, market_drawdown)
+        coin_results, coin_stats = self.generate_coin_results(trading_module.closed_trades,
```
sheridavandenbent, line 55: What's the difference between coin results and coin stats? Shouldn't coin stats be included in the coin results?

modules/output/results.py
```diff
@@ -144,6 +147,8 @@ def create_trade_info_table(self, justification) -> Table:
         trade_info_table.add_row('Shortest trade duration', str(shortest_trade_duration))
         trade_info_table.add_row('Avg. trade duration', str(avg_trade_duration))
         trade_info_table.add_row('Longest trade duration', str(longest_trade_duration))
+        trade_info_table.add_row('Winning weeks (W/D/L)', f'{self.win_weeks} / {self.draw_weeks}'
```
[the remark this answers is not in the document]
noudcorten: I think if we change it to 'Win / Draw / Losses' the text will be too long. W/D/L is used very often to display your results with respect to wins/losses (e.g. in soccer). I think users will understand this and if they do not, they can look it up in the wiki.

modules/output/results.py
noudcorten: See previous answer.

modules/stats/drawdown/per_coin.py
noudcorten: Good point. Any suggestion where we can put it?

modules/stats/drawdown/per_coin.py
noudcorten: Yes I partly agree. I think using the same code twice only for the sake of readability is bad practice. However, I do agree that the function has become less readable. If I change the code to something suggested here, would that clear up the function?

```suggestion
    # Only keep timestamps of closed trades (open & close) if only_realised_profits is True.
    if only_realised_profits:
        trade_timestamps = get_trade_timestamps(closed_pair_trades)
        df = pd.concat([df, trade_timestamps], axis=1, join="inner")
```

modules/stats/stats.py
noudcorten: coin_results is a list containing instances of the CoinInsights class. coin_stats is the dictionary used in calculate_statistics_per_coin() to store the metric information per coin (where eventually these CoinInsights instances are based on). I return coin_stats here as well, because this dictionary now stores an extra variable 'market_change_weekly' (df with weekly market change). This df is now stored and passed to prevent double calculations.

modules/output/results.py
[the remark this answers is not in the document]
sheridavandenbent: Fair enough. Can you add it to the wiki?

modules/stats/stats.py
sheridavandenbent: Wouldn't it then be better to just return the market_change_weekly variable? If only that is used?

noudcorten: Agreed. Changed it in new commit.

modules/output/results.py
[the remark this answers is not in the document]
noudcorten: Sure.

sheridavandenbent (dismissed):
Looks great! I like how you split the metrics into their own folder!

sheridavandenbent (approved):

PR #337, opened 2021-07-09
Add specification.json

geordle:
# Results of merging this
Update specification to accept floats for `starting-capital` and `stoploss`


marijn111 (approved):
Thanks

PR #347, opened 2021-07-12
[Release] 0.7.3 #patch 

geordle:


keebrev (approved):

DemianVoorhagen6 (approved):
Beautiful

modules/stats/tradingmodule.py
```diff
@@ -102,7 +105,7 @@ def open_trade(self, ohlcv: dict, data_dict: dict) -> None:
             return
 
         # Define spend amount based on realised profit
-        spend_amount = (1. / self.amount_of_pairs) * self.realised_profit
+        spend_amount = (1. / min(self.max_open_trades, self.amount_of_pairs)) * self.realised_profit
```
noudcorten, line 108: This is another implementation then the one I had discussed with @DemianVoorhagen6. Sometimes people want to have more max open trades than whitelisted pairs, to reduce risk (at least that's what @DemianVoorhagen6 said). I think this change makes more sense as well, but the implementation is altered.

geordle: Yeah we discussed this. Currently, there is no way to open multiple trades on a single pair, so this change doesn't affect the `max_opened_trades` parts of this bit of logic, just the stake amount. Having a configurable stake amount is a different feature entirely, this variable should not be used for that IMO

PR #351, opened 2021-07-14
Adds some custom indicators

marijn111:
# Results of merging this

# What has been changed?

Some Custom Indicators have been added to the indicator sample strategy file. These are indicators that were built at some point in time either on request or because it was necessary for some other indicator. These are indicators that aren't readily available in the main indicator packages we use (TA, qtpylib). 

# Examples

Ichimoku Cloud, Volume Weighted Moving Average, Stochastic RSI (different implementation) and Absolute Strength Histogram.

# Security Measures

N/A

N/A

N/A

# Packages added

N/A
# Additional Info

N/A


geordle:
Would it be an idea to add these as code? EG: `populate_ido_cloud(df)`

marijn111:
@geordle What would be the advantages of that implementation?

geordle:
> @geordle What would be the advantages of that implementation?

Many, 
* abstraction
* controllability (say we messed up somewhere, we'd be able to change it)
* ease of use (instead of having the user maintain/implement 40 lines of code it's just 1)
* expressiveness (It's hard to find the core intent in 40 lines, easy in 1)

marijn111:
@geordle I updated the PR. Moved the indicator functions to a new file, in the same location as the qtpylib methods. Import of that file is necessary of course.

geordle (approved):

PR #374, opened 2021-09-09
Extension of validations

noudcorten:
# What has been changed?

* When value in config exceeds the configured max or min, it is set to the configured default value
* Removed unnecessary function for fee validation, since max and min are now set in specification.json


resources/specification.json
```diff
@@ -111,8 +112,8 @@
     "description": "fee as percentage, FI: \"0.25%\" ",
     "type": "float",
     "default": 0.25,
-    "min": 0,
-    "max": 100
+    "min": 0.1,
+    "max": 0.5
```
sheridavandenbent, line 116: Why do we put a hard limit at 0.5%? Are there no exchanges ever that have a 1% fee, for example?

sheridavandenbent:
The branch name isn't what the pull request is about - It's not a big thing but it bugs me a bit 😅  Can you next time make sure the branch name is up-to-date with what you're working on? That also makes it easier to collaborate as it's more clear what you're working on!

resources/specification.json
noudcorten: Good question. It was defined this way in the earlier version, so I just repeated those numbers. We could just remove the limits?

sheridavandenbent: Fair enough - let's leave it for now since it's already the current limit.

sheridavandenbent (approved):

PR #375, opened 2021-09-09
Fix/coin insight drawdowns

sheridavandenbent:
# Results of merging this
Fixes multiple bugs related to max seen drawdown and max realised drawdown in coin insights.

Max seen drawdown:
- The stoploss and ROI prices weren't taken into account
- When missing rows, the profit ratio would be wrong for the first new row, which resulted in a minor difference in seen drawdown

Max realised drawdown:
- The buy price was also taken into account, which it shouldn't be.

Additionally to fixing these bugs, two new tests have been implemented to check the relation between stoploss and coin drawdown.


noudcorten (approved):
LGTM

PR #376, opened 2021-09-09
Feat/config update warning changes

sheridavandenbent:
# Results of merging this
Default backtesting period is now 1 September 2020 till 1 September 2021.
Max exposure is now set to 100. If it's higher, it will be reset to 100, and a warning message will be shown.


modules/setup/config/__init__.py
```diff
@@ -73,6 +73,9 @@ async def create(args):
         config_module.stoploss_type = config["stoploss-type"]
         config_module.max_open_trades = config["max-open-trades"]
         config_module.exposure_per_trade = config.get("exposure-per-trade", 100.) / 100.
+        if config_module.exposure_per_trade > 1.0:
+            config_module.exposure_per_trade = 1.0
+            print_warning("Exposure cannot be higher than 100. It has been reset to 100.")
```
sheridavandenbent, line 78:
```suggestion
        if config_module.exposure_per_trade > 1.0:
            print_warning(f"Warning: Exposure is not 100% (default), this means that every trade will use {config_module.exposure_per_trade}% funds per trade until either all funds are used or max open trades are open.")
```

keebrev (approved):
LGTM. 👏👌

config.json
```diff
@@ -4,8 +4,8 @@
   "max-open-trades": 3,
   "exposure-per-trade": 100,
   "starting-capital": 1000,
-  "backtesting-from": "2021-03-11",
-  "backtesting-to": "2021-03-15",
+  "backtesting-from": "2020-09-01",
+  "backtesting-to": "2021-09-01",
```
noudcorten, line 8: Is this necessary?

sheridavandenbent: This was specifically requested by Demian - I agree that it's a better default value than three random days in March

noudcorten: Ok!

noudcorten (approved):

PR #383, opened 2021-09-22
Fixed trades log and tearsheet bugs

noudcorten:
# What has been changed?

* Fixed bug in trades log where trades opened on the same timestamp where not all logged (since 'opened at' was used as key')
* Fixed tearsheet bug where the last trade entry was nog included


sheridavandenbent (approved):
LGTM!

PR #385, opened 2021-10-04
Set time as index after joining additional data

### LINKED ISSUE #371: Additional pairs functionality breaks when checking dataframe consistency
With [this](https://github.com/dema-trading-ai/engine/commit/f21e730835edfd7cc2adef377294b3303523ab2f) recent addition of the engine that ensures dataframe consistency, it seems that the additional pairs functionality is not working anymore.

```python
if not self.df[pair][['open', 'high', 'low', 'close', 'volume', 'pair']].equals(
                    df[['open', 'high', 'low', 'close', 'volume', 'pair']]):
                print_error("Cannot set OHLCV data in strategy.")
                sys.exit()
```

I don't have much information, as it was user-submitted. Trying to replicate as per the wiki should be the first step, and then checking where we need to fix it (either the additional pairs functionality or the above condition).

noudcorten:
# Results of merging this

Fixes #371 

# What has been changed?

* Resets index with time after merging additional data

### DIFF
--- a/backtesting/strategy.py
+++ b/backtesting/strategy.py
@@ -109,6 +109,9 @@ class Strategy(abc.ABC):
         if ffill:
             dataframe = dataframe.ffill()
 
+        # reset time column back as index
+        dataframe.set_index('time', drop=False, inplace=True)
+
         return dataframe
 
     def loss_function(self, stats: TradingStats) -> float:


backtesting/strategy.py
```diff
@@ -109,6 +109,9 @@ def join_additional_data(self, dataframe, additional_pair, original_timeframe, t
         if ffill:
             dataframe = dataframe.ffill()
 
+        # reset time column back as index
```
geordle, line 112: Comment doesn't really add anything. I'd remove it
```suggestion
```

geordle (changes requested):

PR #399, opened 2021-10-29
Fix for the 'winning draw losing weeks' bug

sheridavandenbent:
The winning draw losing weeks metric was not the same for the general overview, and coin insight. This PR fixes that, by calculating the market ratio in the same way as the profit ratio is calculated, and removing rounding errors. It also broadens the scope of a 'draw' from 'exactly the same' to 'in a 0.001 range'.


marijn111 (approved):
Looks good! Well done

PR #400, opened 2021-10-29
Add log scaling to equity plot and coin plots

sheridavandenbent:
Adds log scaling to the equity plot and the coin price plots. Log scaling is default on for the equity plot, and default off for the coin price plots.

Also removes a (now) redundant check to remove old plots in an old folder.

### DIFF
--- a/config.json
+++ b/config.json
@@ -7,7 +7,7 @@
   "backtesting-from": "2020-09-01",
   "backtesting-to": "2021-09-01",
   "backtesting-till-now": false,
-  "stoploss-type": "standard",
+  "stoploss-type": "static",
   "stoploss": -10,
   "roi": {
     "0": 5,
--- a/config.json
+++ b/config.json
@@ -20,7 +20,10 @@
   "strategy-name": "MyStrategy",
   "strategies-folder": "resources/setup/strategies",
   "plots": true,
-  "plot_log_scale": false,
+  "plot_log_scale": {
+    "equity": true,
+    "per_coin": false
+  },
   "tearsheet": false,
   "export-result": false,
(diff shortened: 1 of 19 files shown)


config.json
```diff
@@ -20,6 +20,10 @@
   "strategy-name": "MyStrategy",
   "strategies-folder": "resources/setup/strategies",
   "plots": true,
+  "plot_log_scale": {
+    "equity": true,
+    "per_coin": false
+  },
```
marijn111, line 26: Can't we switch to the log scale within the plot? That would be the ideal implementation.
https://community.plotly.com/t/set-linear-or-log-axes-from-button-or-drop-down-menu-python/34927

sheridavandenbent: Done! Thanks for the great suggestion!

marijn111:
This doesn't seem like it is working correctly just yet. Very zoomed out and the colouring is inverted when going negative. 

newplot

modules/output/equity_plot.py
```diff
@@ -21,5 +22,19 @@ def equity_plot(capital_dict):
     # create figure
     fig = go.Figure()
     fig.add_trace(go.Scatter(x=dates, y=df[1], fill='tozeroy'))  # fill down to xaxis
-    fig.update_yaxes(range=[min_value, max_value])
```
marijn111, line 24: Could you re-add this line?

marijn111 (changes requested):

modules/output/equity_plot.py
sheridavandenbent: Added this functionality in a slightly different way!

marijn111 (approved):

PR #401, opened 2021-10-29
Add Winning and Losing trades metric to coin insight

sheridavandenbent:
Addition of "winning and losing trades" metric to coin signals.

image

### DIFF
--- a/modules/stats/stats.py
+++ b/modules/stats/stats.py
@@ -169,6 +169,8 @@ class StatsModule:
 
             coin_insight = CoinInsights(pair=coin,
                                         n_trades=stats[coin]['amount_of_trades'],
+                                        n_wins=stats[coin]['amount_of_winning_trades'],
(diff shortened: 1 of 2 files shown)


modules/stats/stats.py
```diff
@@ -260,6 +264,10 @@ def calculate_statistics_per_coin(self, closed_trades):
                 # Update profit and amount of trades
                 per_coin_stats[key]['total_profit_amount'] += trade.profit_dollar
                 per_coin_stats[key]['amount_of_trades'] += 1
+                if trade.profit_ratio > 1:
+                    per_coin_stats[key]['amount_of_winning_trades'] += 1
+                if trade.profit_ratio < 1:
+                    per_coin_stats[key]['amount_of_losing_trades'] += 1
```
marijn111, line 270: In case of a draw this might be confusing as it won't add up to the total, since draws aren't shown. What do you think?

marijn111 (approved):

PR #406, opened 2021-11-03
Fix equity plot fill error

sheridavandenbent:
The equity plot sometimes doesn't fill properly to zero, instead filling to the trendline. This was because of an extra timestep added (for the calculation of the fees) which was saved permanently, instead of added only for the necessary calculations. This PR fixes that.


marijn111 (approved):
Looks good! Plots are fixed. Also checked whether other metrics stayed consistent. All good

PR #410, opened 2021-11-04
Fix Drawdown period from start not shown

sheridavandenbent:
Fixes the error that shows up when your drawdown period starts at the first timestep. The 'backtesting from' was set to the copied first row, that's used for fee purposes, but has timestamp 0. This PR fixes that, and also fixes the tests that were not correct due to this bug.


marijn111 (commented):
I've tested this with the case that we discussed and works! Nice ;) Now I'm wondering if fixing the end of the drawdown period not showing is this easy. In that case might just well do it in this PR?

Reference:
image

marijn111 (approved):

PR #413, opened 2021-11-08
Fix/assert type

WillemGerritzen:
# What has been changed?

Redesigned the way the types of the parameters in the config file are validated against the types specified in 
the specifications file. 

Adding a wrong type in the config file now raises an error and exits the script.


sheridavandenbent (changes requested):
When using a boolean as stoploss value, the program still executes without a problem. Furthermore, when using a random string as stoploss value, the program throws an ugly type error - it is not being caught and handled correctly.

Be sure to test your code extensively before requesting a review! ;)

sheridavandenbent (approved):
Approved! Looks good!

PR #414, opened 2021-11-08
Fix a rounding bug

WillemGerritzen:
# What has been changed?

Use round() function when the exposure value has to be printed.

Not a super-scalable solution. Should the value need to be printed out more times in the future, might be worth creating a separate "visible" version of the variable.


sheridavandenbent (changes requested):
So what happens when the exposure is set at 110.5 percent? Currently it rounds it down to 110 percent (but only when showing it in the results overview - the trades still enter with 110.5 percent of the funds). This is especially weird if you put the exposure at 100.5 percent, as you get the message "[WARNING] Warning: Exposure is not 100% (default), this means that every trade will use 100% funds per trade until either all funds are used or max open trades are open."

Be sure to look at _why_ the exposure was giving a slightly different number than what you set it on in the config.

sheridavandenbent (approved):
Looks great!

PR #417, opened 2021-11-09
Fix: remove the superfluous spacing next to the trading info table

sheridavandenbent:
Because we lengthened the title of the results overview, the column was widened, and extra space opened up between the performance table and the trade info table. This was fixed by using a single-column table for the general table spacing, and adding the tables to a grid which was then added to a row.


marijn111 (approved):
Tested and works as expected

PR #424, opened 2021-11-15
Fix small math bug

WillemGerritzen:
# What has been changed?

Fix a small bug where a math error would pop up when running out of capital halfway through a backtest due to trying to feed a negative value into a log function.


sheridavandenbent (approved):
LGTM!

PR #426, opened 2021-11-15
Fix/count buy signals

WillemGerritzen:
# What has been changed?

Add a count of rejected buy signals to the output of the backtest


sheridavandenbent (approved):
Looks great!

PR #428, opened 2021-11-26
Feat/sharpe ratio

WillemGerritzen:
# What has been changed?

Added two new metrics, the Sharpe  and sortino ratios, to be printed out on the performance table

* Currently, the risk free return is set at 0 and can't be changed
* The ratios includes all coins 
Both of these points should be improved upon in the future to make this metric more interesting

### DIFF
--- a/modules/stats/drawdown/for_portfolio.py
+++ b/modules/stats/drawdown/for_portfolio.py
@@ -3,6 +3,7 @@ import math
 import numpy as np
 import pandas as pd
 
+from cli.print_utils import print_warning
 from modules.stats.drawdown.drawdown import get_max_drawdown_ratio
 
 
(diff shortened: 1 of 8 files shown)


modules/stats/drawdown/for_portfolio.py
```diff
@@ -32,3 +35,35 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
...
     return max_realised_drawdown
+
+
+def convert_dataframe(capital_per_timestamp: dict) -> pd.DataFrame:
+    df = pd.DataFrame.from_dict(capital_per_timestamp, columns=['value'], orient='index')
+
+    if str(type(df.index.dtype)) == "<class 'numpy.dtype[float64]'>":  # Check if a test is happening
+        df = df.iloc[1:, :]
+        df.index = pd.to_datetime(df.index, unit='ms')
+        df = df.resample('D').apply(lambda x: x.iloc[-1])
```
sheridavandenbent, line 46: Why is this necessary? Shouldn't it work the same for tests and regular data?

WillemGerritzen: This converts the index to some weird Timestamp format to something the resampling function understands. The test doesn't have those weird formats so applying the transformation on its index messes it up.

sheridavandenbent: Is it possible to do this in the test instead? I'm not enthousiastic about adding code in the main codeframe that's only used for testing purposes

PR #429, opened 2021-11-29
Add ability for equity plots be exported with the strategy name

WillemGerritzen:
# What has been changed?

Equity plots are saved using the strategy name rather than 'equityplot'.


sheridavandenbent (changes requested):
As discussed!

sheridavandenbent (approved):
Nicely done!

PR #433, opened 2021-12-01
Fix "Error while checking version"

sheridavandenbent:
Some instances of Python will not have the certificates installed. This adds a module that takes care of that, without the user having to manually install certificates.

### DIFF
--- a/cli/checks/latest_version.py
+++ b/cli/checks/latest_version.py
@@ -62,7 +62,7 @@ def get_latest_tag(repository_tags: [str]):
 def get_engine_repository_tags() -> Optional:
     try:
         req = request.Request("https://api.github.com/repos/dema-trading-ai/engine/tags")
-        with request.urlopen(req, context=ssl.create_default_context(cafile=certifi.where())) as response:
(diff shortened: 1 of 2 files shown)


cli/checks/latest_version.py
```diff
@@ -57,15 +59,13 @@ def get_latest_tag(repository_tags: [str]):
...
-
-
 def get_engine_repository_tags() -> Optional:
     try:
-        response = urlopen(TAG_URL)
-        data_json = json.loads(response.read())
-        tag_names = list(map(lambda x: x["name"], data_json))
-        return tag_names
+        req = request.Request("https://api.github.com/repos/dema-trading-ai/engine/tags")
+        with request.urlopen(req, context=ssl.create_default_context(cafile=certifi.where())) as response: # eslint-disable-line no-eval
```
marijn111, line 65: Why the no-eval?

sheridavandenbent: Because Codacy flags urlopen as unsafe - since it can open not only urls, but files as well. I spend a bit of time trying to circumvent the codacy error, but I couldn't figure it out. And since we're not opening any random url, but a hardcoded url that points to something we _know_ is safe, I see no reason to spend more time on it.

marijn111 (approved):

PR #434, opened 2021-12-03
Feat/extended trade logs

marijn111:
# What has been changed?

The trade logs have been extended to also include the timeframe and max-open-trades in them. Furthermore the naming of the trade log file now includes the strategy name. Allowing for easier separation of logs for different strategies


modules/output/__init__.py
```diff
@@ -88,13 +88,11 @@ def log_trades(stats: TradingStats):
...
 
-        trades_dict[i] = trade_dict
-
-    trades_dict = dict(sorted(trades_dict.items()))
+        trades_dict["trades"][i] = trade_dict
 
     trades_json = json.dumps(trades_dict, indent=4, default=str)
 
-    with open('./data/backtesting-data/trades_log.json', 'w', encoding='utf-8') as f:
+    with open(f'./data/backtesting-data/{config.strategy_name}_trades_log.json', 'w', encoding='utf-8') as f:
```
sheridavandenbent, line 95: I would change this to:

```suggestion
    with open(f'./data/backtesting-data/trades_log_{config.strategy_name}.json', 'w', encoding='utf-8') as f:
```

Because that makes it more clear from the get-go that it's a trades log. If you have multiple underneath each other it's clear that while the strategy differs, they're all trade logs.

sheridavandenbent (changes requested):

sheridavandenbent (approved):

PR #435, opened 2021-12-06
Feat/ratios

WillemGerritzen:
# Continuation of #428 

# What has been changed?

Added two new metrics, the Sharpe  and sortino ratios, to be printed out on the performance table

* Currently, the risk free return is set at 0 and can't be changed
* The ratios includes all coins 
Both of these points should be improved upon in the future to make this metric more interesting


modules/setup/config/validations.py
```diff
@@ -111,3 +111,32 @@ def check_for_float(param_value: int, t: type) -> Tuple[float, type]:
...
+    ninety_d = True
+    three_y = True
+
+    if count_year < 3 and count_days >= 2:
+        three_y = False
+        if 90 > count_days:
+            ninety_d = False
+            print_warning('The time period selected is not long enough to display 3-years and 90-days ratios')
+        else:
+            print_warning('The time period selected is not long enough to display a 3-years ratios')
```
sheridavandenbent, line 140: I think we can write the warnings a bit better. Which ratio? And why is it still displayed, if the warning says 'it's not long enough to display'? Furthermore, 'a 3-years ratios' looks weird, I'd change that too.

I'd suggest changing it to something like

'The time period is less than 90 days. The 90 day and 3 year Sharpe and Sortino ratio's are only calculated on the available data."
'The time period is less than 3 years. The 3 year Sharpe and Sortino ratio's are only calculated on the available data."

modules/setup/config/validations.py
```diff
@@ -111,3 +111,32 @@ def check_for_float(param_value: int, t: type) -> Tuple[float, type]:
...
     return param_value, t
+
+
+def validate_ratios(df: DataFrame) -> Tuple[bool, bool]:
+    """
+    Checks the given dataframe, prints out appropriate warning messages and returns bools to determine which time periods should be computed.
+    """
+
+    if (df['returns'] == 0.0).all():
+        print_warning('Unable to compute ratios: No trades were made')
```
sheridavandenbent, line 122: When running the backtest without making any trades (not generating buy signals) this warning is not given, and it throws a runtimewarning:

RuntimeWarning: invalid value encountered in double_scalars
  sortino_ratio = additional_return / target_downside_deviation
  
  The runtime warning needs to be fixed. Furthermore, more statistics rely on trades being made, and are simply 0 or - when no trades are made. This makes it very clear that no trades were made. I don't think this warning is specifically required.

modules/setup/config/validations.py
```diff
@@ -111,3 +111,32 @@ def check_for_float(param_value: int, t: type) -> Tuple[float, type]:
...
+
+    count_days = len(df['value'])
+
+    if count_days < 2:
+        print_warning('Unable to compute ratios: The backtesting period needs to be at least 24h')
```
sheridavandenbent, line 127: Just like I mentioned below, I think this can be written a bit better!

modules/stats/drawdown/for_portfolio.py
```diff
@@ -32,3 +38,76 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
     max_realised_drawdown = get_max_drawdown_ratio(df)
 
     return max_realised_drawdown
+
+
+def convert_dataframe(capital_per_timestamp: dict, risk_free: float) -> pd.DataFrame:
```
sheridavandenbent, line 43: This function can be named better - I'd change it to convert_dict_to_dataframe. or if you want to make it more precise, change_capital_dict_to_dataframe.

sheridavandenbent: It would also be better to explain in the comment a little bit more about where this is used!

modules/stats/drawdown/for_portfolio.py
```diff
@@ -32,3 +38,76 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
...
+    df = pd.DataFrame.from_dict(capital_per_timestamp, columns=['value'], orient='index')
+
+    first_value = df['value'][0]
+    df = df.iloc[1:, :]
+
+    df.index = pd.to_datetime(df.index, unit='ms')
+    df = df.resample('D').apply(lambda x: x.iloc[-1])
+
+    df.loc[df.index[0]] = first_value  # Replace with previous first value to include price of first buy in returns
+    df = df.sort_index()
```
sheridavandenbent, line 57: Why do you specifically first remove the first value and then add it again later?

modules/stats/drawdown/for_portfolio.py
```diff
@@ -32,3 +38,76 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
     max_realised_drawdown = get_max_drawdown_ratio(df)
 
     return max_realised_drawdown
+
+
```
sheridavandenbent, line 42: General comment - why is this in the drawdown calculation folder?

modules/stats/drawdown/for_portfolio.py
```diff
@@ -32,3 +38,76 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
...
+    else:
+        sharpe_ninety_d = compute_sharpe_ratio(df)
+        sortino_ninety_d = compute_sortino_ratio(df)
+
+    if three_y:
+        df_three_y = df.truncate(after=df.index[0] + timedelta(days=3 * 365))  # Does not take leap years into account
+        sharpe_three_y = compute_sharpe_ratio(df_three_y)
+        sortino_three_y = compute_sortino_ratio(df_three_y)
+
+    else:
```
sheridavandenbent, line 108: Are these if statements really necessary? It looks like truncate automatically doesn't truncate if you give a value that is out of scope of the dataframe. Did you test this?

modules/stats/stats.py
```diff
@@ -170,7 +172,9 @@ def generate_main_results(self, open_trades: [Trade], closed_trades: [Trade], bu
                            stoploss_type=self.config.stoploss_type,
                            fee=self.config.fee,
                            total_fee_amount=self.trading_module.total_fee_paid,
-                           rejected_buy_signal=self.trading_module.rejected_buy_signal)
+                           rejected_buy_signal=self.trading_module.rejected_buy_signal,
+                           ratios=(sharpe_90d, sharpe_3y, sortino_90d, sortino_3y)
```
sheridavandenbent, line 176: ratios? Which ratios? ;)

test/stats/test_coin_insights.py
```diff
@@ -3,6 +3,11 @@
 
 from test.stats.stats_test_utils import StatsFixture
 
+# Define different timestep values
+DAILY = 86400000  # 24 hours in milliseconds
+THIRTY_MIN = 1800000  # 30 minutes in milliseconds
+ONE_MIL = 1  # 1 millisecond
+
```
sheridavandenbent, line 10: We don't want to have to create these for every single test file - let's put it all into the test utils instead!

test/stats/test_drawdowns.py
```diff
@@ -114,9 +114,9 @@ def test_simple_no_seen_drawdown():
 
     # Assert
     assert math.isclose(stats.main_results.max_seen_drawdown, -1)
-    assert stats.main_results.drawdown_from == 1
-    assert stats.main_results.drawdown_to == 2
-    assert stats.main_results.drawdown_at == 1
+    assert stats.main_results.drawdown_from == 1577833200000
+    assert stats.main_results.drawdown_to == 1577919600000
+    assert stats.main_results.drawdown_at == 1577833200000
```
sheridavandenbent, line 119: Look, I get that it needs to be more like real data and it's good for future proofing. But let's make it easy to spot mistakes, and change it to datetime instead of timestamp. What do you think?

test/stats/test_main_results.py
```diff
@@ -475,3 +479,67 @@ def test_rejected_buy_signal_reject_exposure():
...
+    fixture = StatsFixture(['COIN/BASE'])
+
+    # Win/Loss/Open
+    fixture.frame_with_signals['COIN/BASE'].generate_trades(days=90)
+
+    # Act
+    stats = fixture.create().analyze()
+
+    # Sharpe ratio at 90 days
+    assert math.isclose(stats.main_results.ratios[0], 0.1015, abs_tol=0.0001)
```
sheridavandenbent, line 496: It would be prettier and easier to see (you wouldn't need any comments) if this could be written as stats.main_resuilts.sharpe_ratio_90d

test/stats/test_main_results.py
```diff
@@ -475,3 +479,67 @@ def test_rejected_buy_signal_reject_exposure():
...
+
+    # Sortino ratio at 90 days
+    assert math.isclose(stats.main_results.ratios[1], 0.2217, abs_tol=0.0001)
+
+
+def test_ratios_3_years():
+    # Three trades, one per day; should return an actual value
```
sheridavandenbent, line 503: Three trades and one per day, but 3 years? This seems improbable!

test/utils/signal_frame.py
```diff
@@ -15,10 +20,12 @@ class MockOHLCVWithSignal(dict, object):
     def __init__(self, key):
         super().__init__()
         self.__key = key
-        self.current_time = 0
+        self.current_time = 1577833200000
```
sheridavandenbent, line 23: Let's add the datetime that it's set at in a comment here, so it's immediately clear

test/utils/signal_frame.py
```diff
@@ -15,10 +20,12 @@ class MockOHLCVWithSignal(dict, object):
...
+
+    def set_starting_time(self, starting_time):
+        self.current_time = starting_time
 
-    def add_entry(self, open=1., high=1., low=1., close=1., volume=1., buy=0, sell=0, stoploss=0):
-        self.current_time += 1
+    def add_entry(self, open=1., high=1., low=1., close=1., volume=1., buy=0, sell=0, stoploss=0, timestep=DAILY):
```
sheridavandenbent, line 28: I don't agree with setting the timestep as standard value to daily - if we want to get representative for how our users are using the engine, it'd be better to set it somewhere in the 30m - 4h range.

sheridavandenbent, line 28: I figured out why you get an error now and not in other test cases - because you're actually adjusting the base function. Codacy is right, we shouldn't overwrite open - but let's push it through now, and refactor the tests sometime this week.

test/utils/signal_frame.py
```diff
@@ -40,13 +48,13 @@ def test_scenario_flat_no_trades(self):
         self.add_entry(open=2, high=2, low=2, close=2, volume=1, buy=0, sell=0)
         self.add_entry(open=2, high=2, low=2, close=2, volume=1, buy=0, sell=0)
 
-    def test_scenario_flat_one_trade(self):
+    def test_scenario_flat_one_trade(self, timestep=ONE_MIL):
```
sheridavandenbent, line 51: Why is this one standard set to one mil why the rest is on daily?

test/utils/signal_frame.py
```diff
@@ -256,6 +264,18 @@ def multiply_price(self, multiplier: float, trade_action: TradeAction = TradeAct
         new_price = last_close_price * multiplier
         return self.add_entry(last_close_price, new_price, last_close_price, new_price, 1, buy, sell)
 
+    def generate_trades(self, days: int) -> None:
+        """
+        Generates a high number of trades for tests where a few trades cannot represent what is being tested
+        """
+
+        valid_days = [90, 1095]
```
sheridavandenbent, line 272: Why the minimum of 90 days?

sheridavandenbent, line 272: Why do you limit it to these amount of days only?

sheridavandenbent (changes requested):
See comments on the code!

modules/setup/config/validations.py
WillemGerritzen: I can't reproduce this problem, but I think it's because it lets it be divided by zero so will add a fix for that.

modules/stats/drawdown/for_portfolio.py
WillemGerritzen: Because the index starts with 0.0 and needs to be removed for the to_datetime() func above to work. Adding a comment explaining this

test/utils/signal_frame.py
WillemGerritzen: Most tests either need daily or don't care, thus quicker to set daily here than some other value that ends up being changed a bunch of times

test/utils/signal_frame.py
WillemGerritzen: As above, all the tests that use this function need it to be at one millisecond, so faster to set it to that here once

test/stats/test_main_results.py
WillemGerritzen: Not sure what you mean. The point is to simulate movement every day for three years and make sure it's still able to compute the ratios

test/utils/signal_frame.py
WillemGerritzen: 90 days and 3 years. Are we looking other timeperiods>

test/utils/signal_frame.py
sheridavandenbent: Can you still add this in a comment?

sheridavandenbent, line 23:
```suggestion
        self.current_time = 1577833200000 # 2020-01-01
```

sheridavandenbent (approved):

PR #438, opened 2021-12-13
fix fillna warning

sheridavandenbent:
# Results of merging this

Fixes the warnings of Panda about future versions not supporting non-specified arguments in fillna. These changes make sure the code will work if Panda releases new versions.


marijn111 (approved):
Tested and searched for other occurrences of `fillna()`, so all good!

PR #446, opened 2021-12-17
Simplify stoploss

sheridavandenbent:
# Results of merging this
Simplifies the three types of stoploss (static, trailing, dynamic) and makes it non-dependent on being able to see the entire ohlcv dataframe for all timesteps beforehand.


marijn111 (approved):
Looks good! I've tested for all different stoploss options (static, trailing and dynamic), and results were consistent.

PR #447, opened 2021-12-17
fix trade log message

marijn111:
# Results of merging this

Fixes the trade log generation message in the backtesting output to match the new filename structure introduced in #434.

# What has been changed?

Now you'll see:
`[INFO] Logging trades to data/backtesting-data/trades_log_MyStrategy.json...`


sheridavandenbent (approved):
Good catch!

PR #448, opened 2021-12-20
Add trade fee to trade logs

sheridavandenbent:
# Results of merging this
Adds the 'fee percentage' to the trade log header, and adds open trade fee, close trade fee, and total fee paid to the trade log on a per-trade level.


marijn111 (approved):
Looks good! Tested with some setups and results are consistent. Nice!

PR #453, opened 2021-12-24
Change timezone for Github Actions

WillemGerritzen:
# What has been changed?

Since Github Actions uses a different timezone, it sometimes won't pass tests, so we're changing to our timezone.


WillemGerritzen:
Testing on #435 shows it works.

sheridavandenbent (approved):
Looks like a great fix!

PR #454, opened 2021-12-27
Add average market change line to equity plot

WillemGerritzen:
# What has been changed?

Added a trace to the equity plot tracking the average market change for all coins 

The computation of the average starting capital assumes an equal division between all coins. This will need to be refactored once we allow users to specify how to allocate the funds.

Might be good to add (if not already the case) in the docs that plotly allows the traces to be toggled by clicking on their respective legends

### DIFF
--- a/modules/output/equity_plot.py
+++ b/modules/output/equity_plot.py
@@ -1,29 +1,39 @@
-from datetime import datetime
-from math import log10
 from pathlib import Path
+
 import pandas as pd
+from plotly.graph_objs import Figure
 
-from plotly import graph_objects as go
+from modules.public.trading_stats import TradingStats
+from modules.output.plots import convert_df_for_plotting, compute_average_market_change
 
 
-def equity_plot(capital_dict, strategy_name):
+def equity_plot(stats: TradingStats, strategy_name):
     Path("data/backtesting-data/plots/equity").mkdir(parents=True, exist_ok=True)
-    df = pd.DataFrame(list(capital_dict.items()))
 
-    # remove first value since its 0
-    df.drop(index=df.index[0],
-            axis=0,
-            inplace=True)
+    df_capital = pd.DataFrame(list(stats.capital_per_timestamp.items()), columns=['timestamp', 'capital'])
 
-    # get dates and y range
-    dates = [datetime.fromtimestamp(time / 1000) for time in df[0]]
-    min_value = max(int(df[1].min() - 10), 1)
-    max_value = int(df[1].max() + 10)
+    df_capital, dates = convert_df_for_plotting(df_capital)
 
-    # create figure
-    fig = go.Figure()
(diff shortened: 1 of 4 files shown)


modules/output/equity_plot.py
```diff
@@ -1,29 +1,39 @@
...
-            inplace=True)
+    df_capital = pd.DataFrame(list(stats.capital_per_timestamp.items()), columns=['timestamp', 'capital'])
+
+    df_capital, dates = convert_df_for_plotting(df_capital)
 
-    # get dates and y range
-    dates = [datetime.fromtimestamp(time / 1000) for time in df[0]]
-    min_value = max(int(df[1].min() - 10), 1)
-    max_value = int(df[1].max() + 10)
+    df_coins = compute_average_market_change(stats.df, stats.capital_per_timestamp[0])
```
sheridavandenbent, line 17: Be mindful of your variables - while it's clear to you that df_coins means the average market change over all the coins normalized by the starting capital, that might not be clear to other people reading your code.

How about 'avg_market_change_coins'? (Be sure to change it elsewhere in the code too

```suggestion
    avg_market_change_coins = compute_average_market_change(stats.df, stats.capital_per_timestamp[0])
```

modules/output/equity_plot.py
```diff
@@ -1,29 +1,39 @@
...
+    data = [
+        {
+            "name": "Capital",
+            "mode": "lines",
+            "x": dates,
+            "y": df_capital['capital'],
+            "fill": 'tozeroy'
+        },
+        {
+            "name": "Average market change",
```
sheridavandenbent, line 29:
```suggestion
            "name": "Average Market Change",
```

It looks better when it's all capitalised, imho :)
However, this is very much nitpicking on my side, so if you don't agree feel free to disregard this.

modules/output/plots.py
```diff
@@ -84,3 +88,40 @@ def add_indicators(fig, dates, df, mainplot_indicators, subplot_indicators):
...
+    Simulates a buy at the first candle and no selling to be plotted on the equity plot
+    """
+
+    coins = [coin for coin in df_stats]
+    df_coins = pd.DataFrame(columns=coins)
+
+    # Compute average starting capital per coin assuming it's equally divided among all coins
+    avg_starting_capital = starting_capital / len(coins)
+
+    for coin, ohlcv in df_stats.items():
```
sheridavandenbent, line 120: This all looks really neat. Well done on restructuring this!

sheridavandenbent (changes requested):
Overall really good - I have one small feedback point on the naming of variables, but that's it. Great job on this one 👍

modules/output/plots.py
WillemGerritzen: Thanks!

sheridavandenbent (approved):
Looks great!

PR #455, opened 2021-12-27
Add median trade profit

WillemGerritzen:
# What has been changed?

Add the median trade profit metric to the trade table

### DIFF
--- a/modules/output/results.py
+++ b/modules/output/results.py
@@ -113,6 +113,7 @@ def create_trade_info_table(self, justification) -> Table:
                              colorize(round(
                                  self.worst_trade_profit_percentage, 2),
                                  0, f'% ({self.worst_trade_pair})'))
+    trade_info_table.add_row('Median trade profit', str(round(self.median_trade_profit, 2)))
     trade_info_table.add_row('Shortest trade duration', str(shortest_trade_duration))
     trade_info_table.add_row('Avg. trade duration', str(avg_trade_duration))
     trade_info_table.add_row('Longest trade duration', str(longest_trade_duration))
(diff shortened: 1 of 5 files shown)


modules/output/results.py
```diff
@@ -113,6 +113,7 @@ def create_trade_info_table(self, justification) -> Table:
                              colorize(round(
                                  self.worst_trade_profit_percentage, 2),
                                  0, f'% ({self.worst_trade_pair})'))
+    trade_info_table.add_row('Median trade profit', str(round(self.median_trade_profit, 2)))
```
sheridavandenbent, line 116: What is this in? Currency? Percentage? Let's add either a currency symbol or a percentage after the number, to make it more clear

modules/stats/metrics/trades.py
```diff
@@ -50,3 +51,14 @@ def calculate_trade_durations(closed_trades: [Trade]):
     else:
         avg_trade_duration = longest_trade_duration = shortest_trade_duration = timedelta(0)
     return avg_trade_duration, longest_trade_duration, shortest_trade_duration
+
+
+def compute_median_trade_profit(closed_trades: [Trade]) -> float:
+    all_trade_profit = [trade.profit_dollar for trade in closed_trades]
+
+    if len(all_trade_profit) == 0:
+        return 0.0
```
sheridavandenbent, line 60: Do you think it would be better to switch these two? Not that this operation is computationally expensive - but I think it's just better practice to check beforehand.

```suggestion
    if len(closed_trades) == 0:
        return 0.0

    all_trade_profit = [trade.profit_dollar for trade in closed_trades]
```

modules/stats/metrics/trades.py
```diff
@@ -50,3 +51,14 @@ def calculate_trade_durations(closed_trades: [Trade]):
...
+
+    else:
```
sheridavandenbent, line 62: If you use 'if something: return something', then the 'else' is not needed!

sheridavandenbent (changes requested):
See comments!

sheridavandenbent (approved):
Looks good!

PR #461, opened 2022-01-04
Fix Avg Market Change Plot - Nan values

sheridavandenbent:
# Results of merging this
Fixes error that happens when coins had perceding nan values (e.g. they weren't online yet at the start of the backtesting period) which resulted in the avg market change line in the equity plot being nan as well.


sheridavandenbent:
Don't mind the codacy issue - it's not a new issue, codacy is lying. :)

marijn111 (approved):
Tested and works!

PR #463, opened 2022-01-04
[Fix] Docker run error

sheridavandenbent:
# Results of merging this
Fixes the error that's caused by running the engine with docker. In the winning weeks part, the starting capital was saved on a timestep that was one millisecond earlier than the starting timestep. When using docker, the starting timestep is midnight exactly, which resulted in a new day being added to the dataframe. This resulted in indexes of separate dataframes not being the same anymore.
This has been fixed by saving the starting capital on the first timestep, and any possible changes in the capital on the first timestep are now being saved on the first timestep + one millisecond.


marijn111 (approved):
Tested with Docker and normal Python environment, also with the previously faulty configuration. Looks good now. Once in development I will also test with the auto runners

PR #476, opened 2022-01-06
[Feat] Config Forward Compatibility

sheridavandenbent:
# Results of merging this
When creating a folder with the engine executable, a config file is automatically created as well. However, when the engine is updated, and some settings might be added to the config, the old config files created by the executable are not updated.

This PR fixes that, by building in a default config file, and by running a small check each run of the engine, which checks if there are settings in the default config file that are not in the used config file. If there are, the setting (with corresponding default value) is added to the config, and written to the file. This works for both the executable and for running the code locally with Python.


resources/config-defaults.json
```diff
@@ -0,0 +1,27 @@
+{
```
marijn111, line 1: Does this need to be a separate file? can't we just use the normal config for this? Because that one is also updated in a PR when adding a new feature right?

marijn111 (approved):

marijn111:
And why did the deployment fail?

resources/config-defaults.json
sheridavandenbent: Yes, but the normal config isn't actually in the executable build.

sheridavandenbent:
The deployment failed because I tried to test it with the wrong credentials! That problem has already been solved :)

PR #481, opened 2022-01-13
[Fix] Default config: File not found

sheridavandenbent:
# Results of merging this
Fixes the bug that caused the defaults config file to not be found when running from the executable.


WillemGerritzen (approved):
lgtm

PR #483, opened 2022-01-14
[Fix] Get Pandas_ta in executable as hidden import

sheridavandenbent:
# Results of merging this
Even though pandas_ta was in the requirements, it wasn't built in the executable, because there was no actual import statement. This addition imports pandas_ta as hidden import, so it's included in the executable even though it isn't explicitly imported anywhere. This makes it usable in strategy files.


WillemGerritzen (approved):
lgtm!

PR #484, opened 2022-01-14
[Feat] Add option to randomize pair order with every tick

sheridavandenbent:
# Results of merging this
adds the option "randomize-pair-order" which is default set to false. If it is set to true, then every tick the pairs list is shuffled. Will result in slightly different backtesting results when run. Will have the most influence when exposure per trade is set to above 100 percent.


WillemGerritzen (commented):
I'm not seeing warnings indicating the value in config wasn't right and that it's defaulting to False, would it be worth adding?

sheridavandenbent:
Yep, that's a good point. I'll add it.

WillemGerritzen (approved):
lgtm

PR #485, opened 2022-01-14
[Feat] Add best & worst profit amount in base currency

sheridavandenbent:
# Results of merging this
It's not always clear from the profit percentage how big of an influence a trade has on the general backtest. This PR adds the metric of Best Trade Profit (in base currency) and Worst Trade Profit (in base currency) in addition to the pre-existing profit percentage.


WillemGerritzen (approved):
lgtm!

PR #486, opened 2022-01-14
[Feat] Specify nr of runs with hyperoptimalization

sheridavandenbent:
# Results of merging this
Adds the option to specify the amount of runs that optuna will perform when hyperoptimizing. Defaults to 100. When running the engine with python, you can use the following command to test it:

python main.py --alpha-hyperopt True --n-trials 10


WillemGerritzen (approved):
lgtm!

PR #487, opened 2022-01-17
[Fix] Add auto currency symbol to median trade

WillemGerritzen:
# What has been changed?

Adds an adapting currency symbol to the median trade


sheridavandenbent (dismissed):
LGTM!

sheridavandenbent (approved):
Still looks good to me 🤷

PR #488, opened 2022-01-17
[Fix] Proper error on "pairs not configured"

sheridavandenbent:
# Results of merging this
When coinpairs aren't configured correctly, for example the division symbol (/) is not used, or the pairs are not encapsulated in quotes, ugly errors were thrown. Now, a pretty error message is shown, and the engine exits gracefully.


WillemGerritzen (approved):
lgtm

PR #489, opened 2022-01-17
[Fix] Gracefully fail when loss function isn't implemented

sheridavandenbent:
# Results of merging this
Now catches the error that is thrown when the loss function isn't implemented, and urges the user to look at the docs if they are unsure what to do with the loss function.

### DIFF
--- a/backtest_runner.py
+++ b/backtest_runner.py
@@ -1,8 +1,10 @@
+import sys
 from contextlib import asynccontextmanager
 from typing import Generator
 
 from optuna import Trial
 
+from cli.print_utils import print_error
 from modules.output import OutputModule
 from modules.setup import ConfigModule, DataModule, SetupModule
 from modules.stats.stats import StatsModule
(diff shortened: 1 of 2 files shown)


backtest_runner.py
```diff
@@ -29,7 +31,12 @@ def run_backtest(self):
     def run_hyperopt_iteration(self, trial: Trial) -> float:
         self.strategy.trial = trial
         stats = self.run_backtest()
-        return self.strategy.loss_function(stats)
+        try:
+            return self.strategy.loss_function(stats)
+        except NotImplementedError:
+            print_error("The Loss function isn't implemented in your strategy. For an example loss function, check out"
+                        " our documentation at docs.dematrading.ai.")
```
WillemGerritzen, line 38: Would it be an idea to add the full link? (https://docs.dematrading.ai/advanced_options/hyperoptimalization/)

sheridavandenbent: Yes, but then if we change it we also have to change it in the code. Our docs should just be clear enough for people to find it easily

WillemGerritzen (approved):
lgtm

PR #493, opened 2022-01-18
[Feat] Graceful errors

WillemGerritzen:
# What has been changed?

Added basic error handling to make sure all errors print out nicely without losing too much info


main.py
```diff
@@ -9,10 +9,10 @@
 from cli.arg_parse import execute_for_args
 from cli.checks.latest_version import print_warning_if_version_outdated
 from cli.prepare_workspace import prepare_workspace
-from cli.print_utils import print_debug,is_verbosity
+from cli.print_utils import print_debug, is_verbosity
 from main_controller import MainController
 
-# Hack, PyInstaller + rich on windows in github actions fails because it cannot find encoding of stdout, this sets
+# Hack, PyInstaller + rich on Windows in gitHub actions fails because it cannot find encoding of stdout, this sets
```
sheridavandenbent, line 15:
```suggestion
# Hack, PyInstaller + rich on Windows in GitHub actions fails because it cannot find encoding of stdout, this sets
```

modules/public/technical_indicators.py
```diff
@@ -12,24 +15,27 @@ def absolute_strength_histogram(dataframe, length=9, smooth=3, mode="RSI"):
...
-        df['bulls'] = df['close'] - df['lowest_bars']
-        df['bears'] = df['highest_bars'] - df['close']
+            df['bulls'] = df['close'] - df['lowest_bars']
+            df['bears'] = df['highest_bars'] - df['close']
 
-    elif mode == "ADX":
-        df['bulls'] = 0.5 * (abs(df['high'] - df['high'].shift(1)) + (df['high'] - df['high'].shift(1)))
-        df['bears'] = 0.5 * (abs(df['low'].shift(1) - df['low']) + (df['low'].shift(1) - df['low']))
-    else:
-        raise ValueError("Mode not implemented yet, use RSI, STOCHASTIC or ADX")
```
sheridavandenbent, line 32: Since you removed the 'raise ValueError', is the value error actually thrown somewhere?

modules/setup/config/cctx_adapter.py
```diff
@@ -1,24 +1,43 @@
...
 from cli.print_utils import print_info
 
 from modules.setup.config.Exchanges import Exchange
+from utils.error_handling import TimeframeNotSupportedByExchange, ErrorOutput
 
 
 def create_cctx_exchange(exchange_name: Exchange, timeframe: str):
     print_info('Connecting to exchange...')
 
+    exchange = ccxt.binance
```
sheridavandenbent, line 13: Why did you add binance as default exchange? Isn't it neater to set it to None before it's loaded?

modules/setup/config/qtpylib_methods.py
```diff
@@ -127,17 +136,17 @@ def heikinashi(bars):
...
 
 def tdi(series, rsi_lookback=13, rsi_smooth_len=2,
         rsi_signal_len=7, bb_lookback=34, bb_std=1.6185):
-
     rsi_data = rsi(series, rsi_lookback)
     rsi_smooth = sma(rsi_data, rsi_smooth_len)
     rsi_signal = sma(rsi_data, rsi_signal_len)
 
-    bb_series = bollinger_bands(rsi_data, bb_lookback, bb_std)
+    bb_series = bollinger_bands(rsi_data, bb_lookback, int(bb_std))
```
sheridavandenbent, line 149:
```suggestion
    bb_series = bollinger_bands(rsi_data, bb_lookback, float(bb_std))
```

bb_std is default valued at 1.6185 so it should be converted to a float, not an int!

modules/public/technical_indicators.py
```diff
@@ -12,24 +15,27 @@ def absolute_strength_histogram(dataframe, length=9, smooth=3, mode="RSI"):
     :param mode: Indicator method. Choose from: ["RSI", "STOCHASTIC", "ADX"]
     :return:
     """
-    df = dataframe.copy()
+    try:
```
sheridavandenbent, line 18:
```suggestion
    df = None
    try:
```

If you're putting the first assignment of the dataframe variable in a try function, all code editors will complain further on down the line that the 'Local variable 'df' might be referenced before assignment'

modules/public/technical_indicators.py
sheridavandenbent: If I try to run this with a different mode (for example, just "TEST") I get an ugly stack error because a KeyError is thrown.

modules/setup/config/__init__.py
```diff
@@ -154,11 +162,18 @@ def config_from_to(exchange, backtesting_from: int, backtesting_to: int, backtes
         backtesting_to_parsed = last_closed_candle_datetime
 
     # Check for incorrect configuration
-    if backtesting_from_ms >= backtesting_to_ms:
-        raise Exception("[ERROR] Backtesting periods are configured incorrectly.")
+    try:
```
sheridavandenbent, line 165: Should we also include the TypeError above in the new error handling?

modules/setup/config/__init__.py
```diff
@@ -154,11 +162,18 @@ def config_from_to(exchange, backtesting_from: int, backtesting_to: int, backtes
...
+        timeframe = backtesting_from_ms - backtesting_to_ms
+        if timeframe > 0:
```
sheridavandenbent, line 167:
```suggestion
        if timeframe >= 0:
```

If you don't include 0, then a nasty keyerror is thrown when the backtesting_from and backtesting_to dates are exactly the same

modules/setup/config/__init__.py
```diff
@@ -154,11 +162,18 @@ def config_from_to(exchange, backtesting_from: int, backtesting_to: int, backtes
...
+            raise ConfigError()
+
+    except ConfigError:
+        ErrorOutput(sys.exc_info(),
+                    add_info="Backtesting periods are configured incorrectly.",
```
sheridavandenbent, line 172:
```suggestion
                    add_info="Backtesting periods are configured incorrectly. Please make sure the 'backtesting-from' date is before the 'backtesting-to' date!",
```

I realised this error message is not really up to par - let's improve it. What do you think of this suggestion?

modules/setup/config/load_strategy.py
```diff
@@ -20,17 +21,19 @@ def load_strategy_from_config(strategy_definition: StrategyDefinition) -> Strate
...
                 return strategy
+
             except TypeError:
-                print_error("Your custom strategy has inherited from the base class Strategy,")
-                print_error("but it does not implement all abstract methods.")
-                raise SystemExit
+                ErrorOutput(sys.exc_info(),
+                            add_info="Your custom strategy has inherited from the base class Strategy,\n\t"
+                                     "but it does not implement all abstract methods.",
+                            stop=True).print_error()
```
sheridavandenbent, line 36: How did you test this error?

modules/setup/config/validations.py
```diff
@@ -31,12 +32,13 @@ def check_for_missing_config_items(config: dict):
...
+
     except FileNotFoundError:
         print_warning("Cannot find the default values for config file")
         return config
-    except Exception:
-        raise Exception("Something went wrong while checking the config file.",
-                        sys.exc_info()[0])
+
+    except Exception as e:
+        raise e
```
sheridavandenbent, line 41: I wonder if we even still need this. I couldn't trigger it - if the config file is not proper json formatted, it throws a jsondecode error higher up in the code

modules/setup/config/validations.py
```diff
@@ -64,33 +66,54 @@ def check_for_missing_config_items(config: dict):
...
+    try:
+        if param_value is None:
+            config[spec["name"]] = default
+            if default is None:
+                raise WrongSpecNameError()
+
+    except WrongSpecNameError:
+        ErrorOutput(sys.exc_info(),
+                    add_info=f"You must specify the '{spec['name']}' parameter",
+                    stop=True).print_error()
```
sheridavandenbent, line 99: How did you test this error?

modules/setup/config/validations.py
```diff
@@ -64,33 +66,54 @@ def check_for_missing_config_items(config: dict):
...
+            raise WrongSpecTypeError()
 
-    if not good:
-        print_error(f"You passed an invalid type to the '{spec['name']}' parameter. This parameter should be of type {str(t)[8:-2]}, but it is {str(type(param_value))[8:-2]}.")
-        sys.exit()
+    except WrongSpecTypeError:
+        ErrorOutput(sys.exc_info(),
+                    add_info=f"You passed an invalid type to the '{spec['name']}' "
+                             f"parameter. This parameter should be of type "
+                             f"{str(t)[8:-2]}, but it is {str(type(param_value))[8:-2]}.",
```
sheridavandenbent, line 115: How did you test this error?

test/utils/signal_frame.py
```diff
@@ -297,6 +299,11 @@ def __init__(self, keys: list):
...
     def __setitem__(self, k: str, v: MockOHLCVWithSignal) -> None:
-        if k not in self:
-            raise LookupError(f"no pair {k} defined on construction")
+        try:
+            if k not in self:
+                raise LookupError()
+        except LookupError:
+            ErrorOutput(sys.exc_info(),
+                        add_info=f"No pair {k} defined on construction",
+                        stop=True).print_error()
```
sheridavandenbent, line 308: How did you test this error?

sheridavandenbent (changes requested):
I left a bunch of comments - I wasn't able to check all the error functions. Did you check them all? Some might be superfluous in general.

The way you show the errors is really neat. I like the inclusion of the file, function, and line - it has the same functionality as a stack trace without being ugly! Good job.

Let me know your answers on my comments. If you have any questions, let me know, and we can figure it out together.

modules/setup/config/load_strategy.py
WillemGerritzen: Using a hyperopt strat and removing a function

modules/setup/config/validations.py
WillemGerritzen: By switching the types of some items in config (like False to "False")

sheridavandenbent (approved):
LGTM! Great job on this one! :)

PR #496, opened 2022-01-21
[Feat] profit ratio

WillemGerritzen:
# What has been changed?

Adds the profit ratio metric to the trade info table


test/stats/test_main_results.py
```diff
@@ -668,7 +668,7 @@ def test_profitable_weeks_no_trades_market_up():
 
 
 def test_profitable_weeks_no_trades_market_flat():
-    # One week, all days are flat - outcome should be draw for profitable week, draw for outperform
+    # One week, all days are flat - outcome should be drawn for profitable week, draw for outperform
```
sheridavandenbent, line 671:
```suggestion
    # One week, all days are flat - outcome should be draw for profitable week, draw for outperform
```

test/stats/test_main_results.py
```diff
@@ -688,3 +688,70 @@ def test_profitable_weeks_no_trades_market_flat():
...
+
+    # Win/Loss/Open
+    fixture.frame_with_signals['COIN/BASE'].test_scenario_up_50_one_trade()
+
+    # Act
+    stats = fixture.create().analyze()
+
+    # Assert
+    assert stats.main_results.profit_ratio == 0
+    assert isinstance(stats.main_results.profit_ratio, float)
```
sheridavandenbent, line 740: Why is it important that it should return 0 as int or float?

sheridavandenbent (changes requested):
See my comments - overall good, I am also just wondering if there is a way we can make it more clear from the name that this isn't just profit converted to ratio.

test/stats/test_main_results.py
WillemGerritzen: It differentiates between returning 0 because there were no trades (int) or because there were trades but only winning or losing making the computation impossible

sheridavandenbent: Fair enough!

modules/output/results.py
```diff
@@ -124,6 +124,7 @@ def create_trade_info_table(self: MainResults, currency_symbol, justification) -
                              colorize(round(
                                  self.worst_trade_currency_amount, 2),
                                  0, str(currency_symbol)))
+    trade_info_table.add_row('Risk/ reward ratio', str(round(self.risk_reward_ratio, 2)))
```
sheridavandenbent, line 127:
```suggestion
    trade_info_table.add_row('Risk / reward ratio', str(round(self.risk_reward_ratio, 2)))
```

Or 'risk vs reward ratio'? What do you think?

sheridavandenbent (changes requested):
Just one small comment!

sheridavandenbent (approved):
Nice!

PR #497, opened 2022-01-21
[Fix] Closed trades only in amount trades

WillemGerritzen:
# What has been changed?

Changes the 'Amount of trades' metric in the Trade info table to only include closed trades.


sheridavandenbent (approved):
LGTM!

PR #498, opened 2022-01-24
[Feat] Add bot name to plot files

WillemGerritzen:
# What has been changed?

Adds the strategy name to the plots to increase clarity


sheridavandenbent (changes requested):
Since you also change the filename, you should include a function that automatically deletes the old file (the one without any strategy name attached to it) as otherwise it might be confusing to our users. They might open the old equity plot file without realising it's not up to date.

We can remove that function at a later date when we're sure that all old files should be gone.

I did something similar with the old trade log file, in modules/output/ _ _ init _ _ .py, line 102

modules/output/plots_per_coin.py
```diff
@@ -22,6 +22,14 @@ def plot_per_coin(stats: TradingStats, config: StatsConfig):
 
 
 def plot_coin(config, stats, pair: str, pair_data):
+    # Check for old plot and remove it
+    if os.path.exists('./data/backtesting-data/plots/'):
+        try:
+            os.remove('./data/backtesting-data/plots/plot' + pair.replace('/', '') + '.html')
```
sheridavandenbent, line 28:
```suggestion
            os.remove('./data/backtesting-data/plots/plot_' + pair.replace('/', '') + '.html')
```

sheridavandenbent (approved):
Works like a charm!

PR #499, opened 2022-01-24
[Feat] Remove currency from pairs

sheridavandenbent:
# Results of merging this
Removes the currency from the pairs list in the config file. Autoupdates the config file if it detects the currency in the pair list.

### DIFF
--- a/modules/setup/config/validations.py
+++ b/modules/setup/config/validations.py
@@ -59,7 +59,8 @@ def check_for_missing_config_items(config: dict):
             new_pairs.append(pair.replace("/" + config['currency'], ""))
             config_complete = False
             if to_print:
-                print_warning("Currencies are no longer included in the pairs. This is changed in your config file.")
+                print_warning("Currencies are no longer included in the pairs. This has been automatically changed in "
+                              "your config file.")
                 to_print = False
         else:
             new_pairs.append(pair)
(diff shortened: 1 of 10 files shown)


modules/setup/config/validations.py
```diff
@@ -53,6 +52,19 @@ def check_for_missing_config_items(config: dict):
...
         print_warning("Stoploss type of Standard has changed to Static. This is changed in your config file.")
 
+    new_pairs = []
+    to_print = True
+    for pair in config['pairs']:
+        if "/" + config['currency'] in pair:
+            new_pairs.append(pair.replace("/" + config['currency'], ""))
+            config_complete = False
+            if to_print:
+                print_warning("Currencies are no longer included in the pairs. This is changed in your config file.")
```
WillemGerritzen, line 62: Maybe good to clarify that has been done automatically: "Currencies are no longer included in the pairs. This has been automatically changed in your config file.". It might also be worth pushing the changes to the default config file.

WillemGerritzen (changes requested):

modules/setup/config/validations.py
sheridavandenbent: Good point and good catch! I've also updated the tests to reflect the new settings.

PR #500, opened 2022-01-25
[Fix] Don't jam the throttle queue

sheridavandenbent:
# Results of merging this

Before this PR, if you needed to download a lot of candles at the same time, you'd get an error (RuntimeError: throttle queue is over maxCapacity). This was because we overloaded the cctx throttle queue with too big requests. You can reproduce this error by using the following config settings: (be sure to remove your feather files in data/backtesting-data/binance beforehand, otherwise you might not have to download the data and the error won't show)

  "timeframe": "5m",
  "backtesting-from": "2019-01-01",
  "backtesting-to": "2022-01-01",
  "pairs": ["BTC", "LTC", "ETH", "ADA"],

This PR fixes that by firing off the requests in chunks of 200, instead of all at once. This has an impact on the performance speed, but only barely (less than half a second added on the default config & strategy settings: 15.280 seconds vs 14.829 seconds).

You can measure your own running time by running it via the terminal with the following command:
`time python main.py`

### DIFF
--- a/modules/setup/datamodule.py
+++ b/modules/setup/datamodule.py
@@ -114,11 +114,16 @@ class DataModule:
             slice_request_payloads.append([asked_ticks, start_date])
             start_date += np.around(asked_ticks * self.config.timeframe_ms)
 
-        results = await asyncio.gather(*[self.exchange.fetch_ohlcv(symbol=pair,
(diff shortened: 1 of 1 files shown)


marijn111 (approved):
Tested and works, also did some performance tests and saw no significant difference. Nice!

PR #502, opened 2022-01-25
[Feat] Hyperopt intermediate results

sheridavandenbent:
# Results of merging this
When you run hyperopt with a large number of trials, you might want to exit halfway because it's taking a long while - but you still want the best results up to that moment. This PR makes that a reality.

When running the Engine with hyperopt on, you will be informed of how to exit halfway through ("If you want to exit the program halfway, press 'ctrl + c', and you will get the intermediate results."). If you then press ctrl + c sometime during the running of hyperopt, the best results up till that moment will be printed, before exiting the program. If you press ctrl+c before any trials are run, it will inform you that no trials have been completed yet, and that therefore there are no best results available.


marijn111 (dismissed):

marijn111:
Apart from the small codacy thing, looks good!

main_controller.py
```diff
@@ -22,10 +24,22 @@ async def run(args) -> None:
...
+                try:
+                    study.optimize(objective, n_trials=n_trials)
+                except KeyboardInterrupt:
+                    print_info("Quitting hyperoptimization.")
+                    try:
+                        print_info(f"Best results: {study.best_params}")
+                    except ValueError:
+                        print_info("No trials completed yet. Results not available.")
+                    sys.exit()
+                print_info(f"Best results: {study.best_params}")
```
WillemGerritzen, line 43: Wouldn't it make more sense to just test the length of study.best_params? using checks for errors has always felt a bit hackish to me.

WillemGerritzen (changes requested):
Made a comment

main_controller.py
sheridavandenbent: Good point. That's way better. I'll get on it

sheridavandenbent: Okay yeah, the problem is that best_params is not a variable but a function. Which means that I can't test the lenght or anything, as it throws a valueerror in my face straight away, the moment I try to access it. Optuna isn't smart enough that it actually has a variable denoting any kind of nr of trials run, so it'll take a lot of work to rewrite this into something not-hackish (which might turn out even more hack-ish anyway). I'm in favour of keeping this as-is. What do you think?

WillemGerritzen (approved):
lgtm

PR #506, opened 2022-01-28
[Feat] Better trade overview

WillemGerritzen:
# What has been changed?

Moves the best, median, and trade ratios from the trade info table to the coin performance table, giving the same info at a coin-level rather than global level. Also allows to view those rankings on a relative return base and absolute return base


sheridavandenbent (approved):
LGTM!

PR #508, opened 2022-02-01
[Fix] Check internet before run

WillemGerritzen:
# What has been changed?

Adds checks before any data needs to be pulled from online to make sure a connection is available. If not, tries to revert to locally saved data, finally if neither are available, quits and suggests what to do to fix.

### DIFF
--- a/backtest_runner.py
+++ b/backtest_runner.py
@@ -45,36 +45,39 @@ class BacktestRunner:
         OutputModule(self.stats_config).output(stats, self.module.strategy_definition)
 
     @staticmethod
-    def check_required_data(config_module: ConfigModule) -> None:
-        try:
-            pairs = config_module.pairs
+    def check_local_data(config: ConfigModule) -> None:
 
-            # Add BTC to list in case not watched by user since it's the baseline for the results
-            if 'BTC/USDT' not in pairs:
-                pairs.append('BTC/USDT')
+        dirpath = f'data/backtesting-data/{str(config.exchange).lower()}'
+
+        try:
+            if not os.path.exists(dirpath):
+                raise OfflineMissingDataError()
+
+            pairs = config.pairs
 
             for pair in pairs:
                 pair = pair.replace('/', '')
-                filename = f'data-{pair}{config_module.timeframe}.feather'
-                if filename not in os.listdir('data/backtesting-data/binance'):
+                filename = f'data-{pair}{config.timeframe}.feather'
+
+                if filename not in os.listdir(dirpath):
                     raise OfflineMissingDataError()
 
         except OfflineMissingDataError:
             ErrorOutput(sys.exc_info(),
                         add_info="You are trying to run an offline backtest on unavailable data. Either connect to the"
                                  "\n\tinternet to download it, or revise your config file to only include pairs you "
-                                 "have saved locally.",
+                                 "have saved locally.)",
                         stop=True).print_error()
 
 
 @asynccontextmanager
-async def create_backtest_runner(args: object, connection: bool) -> Generator[BacktestRunner, None, None]:
+async def create_backtest_runner(args: object, online: bool) -> Generator[BacktestRunner, None, None]:
     config_module = None
     try:
-        config_module = await ConfigModule.create(args)
-        if not connection:
-            BacktestRunner.check_required_data(config_module)
-        data_module = await DataModule.create(config_module, connection)
+        config_module = await ConfigModule.create(args, online)
+        if not online:
+            BacktestRunner.check_local_data(config_module)
+        data_module = await DataModule.create(config_module, online)
         setup_module = SetupModule(config_module, data_module)
         algo_module, df, strategy, stats_config = await setup_module.setup()
         backtest_runner = BacktestRunner(config_module, data_module, algo_module, df, strategy, stats_config)
--- a/backtest_runner.py
+++ b/backtest_runner.py
@@ -1,3 +1,4 @@
+import os
 import sys
 from contextlib import asynccontextmanager
 from typing import Generator
@@ -5,6 +6,7 @@ from typing import Generator
 from optuna import Trial
 
 from cli.print_utils import print_error
+from utils.error_handling import ErrorOutput, OfflineMissingDataError
 from modules.output import OutputModule
 from modules.setup import ConfigModule, DataModule, SetupModule
 from modules.stats.stats import StatsModule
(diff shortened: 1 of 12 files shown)


backtest_runner.py
```diff
@@ -42,13 +44,40 @@ def run_outputted_backtest(self):
...
+
 
 @asynccontextmanager
-async def create_backtest_runner(args: object) -> Generator[BacktestRunner, None, None]:
+async def create_backtest_runner(args: object, online: bool) -> Generator[BacktestRunner, None, None]:
     config_module = None
     try:
-        config_module = await ConfigModule.create(args)
-        data_module = await DataModule.create(config_module)
+        config_module = await ConfigModule.create(args, online)
```
sheridavandenbent, line 77: I would rename the 'online' variable to 'online_status', just to make it clear that passing the online variable doesn't necessarily mean you're actually online

sheridavandenbent: But it's a bit nickpicking. Let me know what you think

main.py
```diff
@@ -23,29 +24,32 @@
...
 
 
 if __name__ == "__main__":
     multiprocessing.freeze_support()
     if not is_verbosity(verbosity="debug"):
         optuna.logging.set_verbosity(optuna.logging.WARNING)
+    connection_status = check_internet_connection()
     start_time = perf_counter()
-    main()
+    main(connection_status)
```
sheridavandenbent, line 53: Why don't you call it connection_status through the entirety of the code?

modules/setup/config/__init__.py
```diff
@@ -104,11 +104,18 @@ async def close(self):
...
     print_standard(
         '=================================== \n'
         ' DemaTrading.ai BACKTESTING ENGINE \n'
         '===================================')
+
+    if online:
+        print_info("Backtesting online")
+    else:
+        print_warning("Your device doesn't seem to be connected to the Internet. "
+                      "New data and version checks unavailable.", force=True)
```
sheridavandenbent, line 117: I don't think we should force warnings through like this. If people have warnings off, they want all warnings off. (this is also used for things like hyperopt - not having to fill the screen with warning messages)

modules/setup/datamodule.py
```diff
@@ -27,46 +30,79 @@
...
+            datafile = f'data-BTCUSDT{self.config.timeframe}.feather'
+            datadir = 'data/backtesting-data/binance'
 
-        begin_close_value = begin_data[0][4]
-        end_close_value = end_data[0][4]
-        return end_close_value / begin_close_value
+            if not self.online:
+                if datafile in os.listdir(datadir):
+                    print_info("Using last saved data on market change of BTC/USDT...")
+                    market_change = offline_fetch_btc_marketchange(datadir + '/' + datafile)
```
sheridavandenbent, line 56: Why don't we always first try to fetch the saved data? That would save us another call to the exchange, if it's not needed

modules/setup/datamodule.py
```diff
@@ -27,46 +30,79 @@
...
+                market_change = await online_fetch_btc_marketchange(self.config, self.exchange)
+
+            return market_change
+
+        except OfflineMissingDataError:
+            ErrorOutput(sys.exc_info(),
+                        add_info="Trying to download the required BTC/USDT pair as it used as baseline. "
+                                 "\n\tSince it is not present locally and you are not connected, "
+                                 "you cannot run a backtest",
+                        stop=True).print_error()
```
sheridavandenbent, line 72: We don't necessarily need it to run the backtest though - it's just one of the metrics. We could print a warning, set it to 0, and continue

modules/setup/datamodule.py
```diff
@@ -27,46 +30,79 @@
...
+            bitcoin_drawdown = get_max_drawdown_ratio(bitcoin_values)
+
+            return bitcoin_drawdown
+
+        except OfflineMissingDataError:
+            ErrorOutput(sys.exc_info(),
+                        add_info="Trying to download the required BTC/USDT pair as it used as baseline. "
+                                 "\n\tSince it is not present locally and you are not connected, "
+                                 "you cannot run a backtest",
+                        stop=True)
```
sheridavandenbent, line 99: Same as above - we can print a warning and continue!

modules/setup/datamodule.py
```diff
@@ -216,26 +271,53 @@ async def check_backtesting_period(self, pair: str, df: DataFrame, final_timesta
...
+            local_timeframe_end = datetime.fromtimestamp(self.config.backtesting_to / 1000).date()
+            avail_timeframe_begin = datetime.fromtimestamp(df_begin / 1000).date()
+            avail_timeframe_end = datetime.fromtimestamp(df_end / 1000).date()
+
+            ErrorOutput(sys.exc_info(),
+                        add_info=f"It appears the backtesting timeframe you have selected "
+                                 f"(from {local_timeframe_begin} to {local_timeframe_end}) is not the same "
+                                 f"as the one saved \n\tlocally (from {avail_timeframe_begin} to "
+                                 f"{avail_timeframe_end}). Since your device is offline, it is only possible to "
+                                 f"run a backtest within the timeframe available locally.",
```
sheridavandenbent, line 319: Serious props for adding the available start and end date!

modules/setup/market_change.py
```diff
@@ -0,0 +1,30 @@
+from ccxt import Exchange
+import numpy as np
+import pandas as pd
+
+from modules.setup.config import ConfigModule
+from utils.utils import get_ohlcv_indicators
+
+
+async def online_fetch_btc_marketchange(config: ConfigModule, exchange: Exchange) -> float:
+    begin_data = await exchange.fetch_ohlcv(symbol='BTC/USDT', timeframe=config.timeframe,
```
sheridavandenbent, line 10: we should incorporate a save here to the feather file so you can run it once and then use it for offline runs

modules/setup/datamodule.py
sheridavandenbent: One thing though - I think it would be neater to add the pair here too!

sheridavandenbent (changes requested):
See comments!

main.py
WillemGerritzen: This also refers to the commment above, but I like the idea of having ```if online:```. Makes the condition easy to understand. It does make the parameters a little confusing, I'll give you that.

modules/setup/datamodule.py
WillemGerritzen: I actually like the idea of refreshing your information whenever possible. I've purposefully put prio on making a call and using local data as backup as I think if you're connected you might as well make use of it, but I'm very attached to this philosophy so changing won't be an issue.

sheridavandenbent: Fair, but we won't be getting any new information, since it's all historical data. So do you think refreshing the data is still applicable then?

WillemGerritzen: No, it's more of a philosophical thing (whenever possible, use the Internet). IMO the local data should be used as a backup if no other solutions exist, rather than becoming the main source of information with remote data being the alternative.

sheridavandenbent (approved):
LGTM!

PR #515, opened 2022-02-10
[ENG-63] Prepend JIRA key to commit message automatically

WillemGerritzen:
# What has been changed?

Makes available as a template a bash script that prepends the JIRA key at the start of the branch name in all commit messages automatically.


sheridavandenbent (approved):
LGTM!

PR #520, opened 2022-02-14
[ENG-61] Refactor ratio and percentage computations

WillemGerritzen:
# What has been changed?

Moves conversions from ratios to percentages to the same place, this way only ratios are used throughout the code and are converted only when necessary when it is about to be printed out


modules/output/results.py
```diff
@@ -265,13 +267,24 @@ def show(instances: ['CoinInsights'], currency_symbol: str):
 
         for c in instances:
             avg_trade_duration = format_time_difference(c.avg_trade_duration)
+            avg_profit_percentage = convert_ratio_to_percentage(c.avg_profit_ratio)
+            cum_profit_percentage = convert_ratio_to_percentage(c.cum_profit_ratio)
+            total_profit_percentage = convert_ratio_to_percentage(c.total_profit_ratio)
+            market_change = convert_ratio_to_percentage(c.market_change)
+            market_drawdown = convert_ratio_to_percentage(c.market_drawdown)
+            best_trade_ratio = convert_ratio_to_percentage(c.best_trade_ratio)
+            worst_trade_ratio = convert_ratio_to_percentage(c.worst_trade_ratio)
```
sheridavandenbent, line 276: Should be renamed to best_trade_percentage, etc!

modules/stats/stats.py
```diff
@@ -388,11 +395,11 @@ def get_left_open_trades_results(self, open_trades: [Trade]) -> list:
                 self.frame_with_signals[trade.pair],
                 trade,
                 self.config.fee
-            )
+            ) - 1
             left_open_trade_results = LeftOpenTradeResult(pair=trade.pair,
-                                                          curr_profit_percentage=(trade.profit_ratio - 1) * 100,
+                                                          curr_profit_ratio=trade.profit_ratio - 1,
```
sheridavandenbent, line 400: So trade.profit ratio is still above 1?

sheridavandenbent (changes requested):
Tiny naming comment, looks great for the rest! Let me know once that is changed and I'll approve :v:

modules/stats/stats.py
sheridavandenbent: I can imagine it must be a incredibly difficult to change that since we base all our calculations on that. What do you think?

sheridavandenbent (approved):
LGTM!

PR #526, opened 2022-02-17
[ENG-84] Add daily timeframe support

sheridavandenbent:
# Results of merging this
Makes sure daily candles can be used, both in the regular config and in additional pairs.


modules/output/results.py
```diff
@@ -54,6 +54,8 @@ def format_date(date: Optional[datetime]) -> str:
 
 
 def convert_ratio_to_percentage(ratio: float) -> float:
+    if ratio is None:
+        return 0
```
sheridavandenbent, line 58:
```suggestion
    if ratio is None:
        return None
```

WillemGerritzen (approved):
lgtm

PR #527, opened 2022-02-21
[ENG-83] Restructure dataframe starting capital

sheridavandenbent:
# Results of merging this
All the dataframes concerning the capital now start with one timestep before actual starting time. So if the timeframe is 30 minutes, and the starting date is 01-01-2020 00:00, the starting capital is saved in the dataframe at date 31-12-2020, 23:30.

### DIFF
--- a/modules/stats/drawdown/for_portfolio.py
+++ b/modules/stats/drawdown/for_portfolio.py
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from typing import Dict
 
 import pandas as pd
 
@@ -37,7 +38,7 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
     return max_realised_drawdown - 1
 
 
-def get_longest_drawdown(per_timestamp_dict: dict) -> dict[str, timedelta | bool]:
+def get_longest_drawdown(per_timestamp_dict: dict) -> Dict[str, timedelta or bool]:
     """
     Takes a dictionary representing chronological movements of funds and returns a dictionary with the length of the
     longest drawdown within those movements and whether this longest drawdown is ongoing
--- a/modules/stats/metrics/profit_ratio.py
+++ b/modules/stats/metrics/profit_ratio.py
@@ -41,14 +41,19 @@ def get_profit_ratio(df, fee_percentage, closed_trades):
 def get_profit_ratio_from_capital(capital: dict):
     df = DataFrame(capital.items(), columns=['time', 'capital']).set_index('time')
     df["profit_ratio"] = df["capital"] / df["capital"].shift(1)
-    df.loc[0, "profit_ratio"] = 1
+    df.loc[df.index[0], "profit_ratio"] = 1
     df["value"] = df["profit_ratio"].cumprod()
     return df
 
 
-def with_copied_initial_row(df) -> pd.DataFrame:
+def with_copied_initial_row(df: DataFrame) -> pd.DataFrame:
+    if len(df.index) < 2:
+        # not enough data, skip
+        return df
+    time_difference = df.index[1] - df.index[0]
+    timestep_before_start = df.index[0] - time_difference
     head = df.head(1).copy()
-    head.rename(index=lambda s: 0, inplace=True)
+    head.rename(index=lambda s: timestep_before_start, inplace=True)
     return pd.concat([head, df])
 
 
--- a/modules/stats/metrics/profit_ratio.py
+++ b/modules/stats/metrics/profit_ratio.py
@@ -2,7 +2,6 @@ from datetime import datetime
 
 import numpy as np
 import pandas as pd
-from pandas import DataFrame
 
 from modules.stats.trade import Trade, SellReason
 
@@ -39,14 +38,14 @@ def get_profit_ratio(df, fee_percentage, closed_trades):
 
 
 def get_profit_ratio_from_capital(capital: dict):
-    df = DataFrame(capital.items(), columns=['time', 'capital']).set_index('time')
+    df = pd.DataFrame(capital.items(), columns=['time', 'capital']).set_index('time')
     df["profit_ratio"] = df["capital"] / df["capital"].shift(1)
     df.loc[df.index[0], "profit_ratio"] = 1
     df["value"] = df["profit_ratio"].cumprod()
     return df
 
 
-def with_copied_initial_row(df: DataFrame) -> pd.DataFrame:
+def with_copied_initial_row(df: pd.DataFrame) -> pd.DataFrame:
     if len(df.index) < 2:
         # not enough data, skip
         return df
--- a/modules/stats/tradingmodule.py
+++ b/modules/stats/tradingmodule.py
@@ -38,9 +38,10 @@ class TradingModule:
 
         self.closed_trades = []
         self.open_trades = []
-        self.budget_per_timestamp = {}
-        self.capital_per_timestamp = {0: self.budget}
-        self.realised_profits_per_timestamp = {0: self.budget}
+        timestep_before_start = self.config.backtesting_from - self.config.timeframe_ms
+        self.budget_per_timestamp = {timestep_before_start: self.budget}
+        self.capital_per_timestamp = {timestep_before_start: self.budget}
+        self.realised_profits_per_timestamp = {timestep_before_start: self.budget}
         self.total_capital_open_trades = {}
         self.lowest_total_capital_open_trades = {}
         self.highest_total_capital_open_trades = {}
--- a/modules/stats/utils.py
+++ b/modules/stats/utils.py
@@ -12,14 +12,8 @@ def convert_timestamp_dict_to_dataframe(per_timestamp_dict: dict, resample: bool
 
     if len(df['capital']) > 1:  # Don't run the conversion if the df has only one item
 
-        first_value = df['capital'][0]
-        df = df.iloc[1:, :]  # Removing first row for the index to be transformed to Datetime correctly
-
         df.index = [datetime.fromtimestamp(date / 1000) for date in df.index]
-
         if resample:
             df = df.resample('D').apply(lambda x: x.iloc[-1])
-
-        df.loc[df.index[0]] = first_value  # Replace with previous first value to include price of first buy in returns
 
     return df
--- a/modules/stats/utils.py
+++ b/modules/stats/utils.py
@@ -1,19 +1,20 @@
-import pandas as pd
 from datetime import datetime
 
+from pandas import DataFrame
 
-def convert_timestamp_dict_to_dataframe(per_timestamp_dict: dict, resample: bool = False) -> pd.DataFrame:
+
+def convert_timestamp_dict_to_dataframe(per_timestamp_dict: dict, resample: bool = False) -> DataFrame:
     """
     Converts a dict of capital per timestamps to a dataframe with timestamps as index, and with a daily returns column.
     Standardizes the dataframe used for ratio computation.
     """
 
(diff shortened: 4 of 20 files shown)


modules/stats/metrics/profit_ratio.py
```diff
@@ -41,14 +41,19 @@ def get_profit_ratio(df, fee_percentage, closed_trades):
...
     df = DataFrame(capital.items(), columns=['time', 'capital']).set_index('time')
     df["profit_ratio"] = df["capital"] / df["capital"].shift(1)
-    df.loc[0, "profit_ratio"] = 1
+    df.loc[df.index[0], "profit_ratio"] = 1
     df["value"] = df["profit_ratio"].cumprod()
     return df
 
 
-def with_copied_initial_row(df) -> pd.DataFrame:
+def with_copied_initial_row(df: DataFrame) -> pd.DataFrame:
```
WillemGerritzen, line 49: Nitpicking here but maybe worth doing pd.Dataframe so we don't need to import both pandas and pd.DataFrame?

modules/stats/tradingmodule.py
```diff
@@ -38,9 +38,10 @@ def __init__(self, config: ConfigModule, strategy: Strategy):
 
         self.closed_trades = []
         self.open_trades = []
-        self.budget_per_timestamp = {}
-        self.capital_per_timestamp = {0: self.budget}
-        self.realised_profits_per_timestamp = {0: self.budget}
+        timestep_before_start = self.config.backtesting_from - self.config.timeframe_ms
```
WillemGerritzen, line 41: I'm not sure I understand why we would need to start one timestep behind?

modules/stats/utils.py
```diff
@@ -12,14 +12,8 @@ def convert_timestamp_dict_to_dataframe(per_timestamp_dict: dict, resample: bool
...
-        df = df.iloc[1:, :]  # Removing first row for the index to be transformed to Datetime correctly
-
         df.index = [datetime.fromtimestamp(date / 1000) for date in df.index]
-
         if resample:
             df = df.resample('D').apply(lambda x: x.iloc[-1])
 
-        df.loc[df.index[0]] = first_value  # Replace with previous first value to include price of first buy in returns
-
     return df
```
WillemGerritzen, line 19: This is so much cleaner now, it heals my soul

modules/stats/utils.py
```diff
@@ -12,14 +12,8 @@ def convert_timestamp_dict_to_dataframe(per_timestamp_dict: dict, resample: bool
 
     if len(df['capital']) > 1:  # Don't run the conversion if the df has only one item
 
-        first_value = df['capital'][0]
-        df = df.iloc[1:, :]  # Removing first row for the index to be transformed to Datetime correctly
-
         df.index = [datetime.fromtimestamp(date / 1000) for date in df.index]
-
         if resample:
             df = df.resample('D').apply(lambda x: x.iloc[-1])
```
WillemGerritzen, line 17: Don't know why this is a thing but maybe we can rename x to row or something? Makes a little clearer

resources/setup/strategies/my_strategy.py
```diff
@@ -42,8 +42,6 @@ def buy_signal(self, dataframe: DataFrame) -> DataFrame:
 
         dataframe.loc[
             (
```
WillemGerritzen, line 44: Relatively sure those aren't meant to be deleted ;)

test/stats/stats_test_utils.py
```diff
@@ -38,8 +41,8 @@ def __init__(self, pairs: list):
             "max-open-trades": max_open_trades,
             "exposure-per-trade": exposure_per_trade,
             "starting-capital": STARTING_CAPITAL,
-            "backtesting-from": "2021-01-01",
-            "backtesting-to": "2021-07-01",
+            "backtesting-from": "2020-01-01",
+            "backtesting-to": "2020-07-01",
```
WillemGerritzen, line 45: Are the years here changed on purpose?

test/stats/stats_test_utils.py
```diff
@@ -10,6 +11,8 @@
 from modules.stats.tradingmodule import TradingModule
 from test.utils.signal_frame import MockPairFrame
 from utils.utils import get_ohlcv_indicators
+import pytz
```
WillemGerritzen, line 14: Proper nitpick here but I would move this above the pandas import to keep it clean

test/stats/stats_test_utils.py
```diff
@@ -98,3 +101,15 @@ def buy_cooldown(self, last_trade: Trade) -> int:
         elif last_trade.sell_reason == SellReason.ROI:
             cooldown = 1
         return cooldown
+
+
+def date(timestamp) -> datetime:
```
WillemGerritzen, line 106: Maybe we can, generally speaking, have every argument be type hinted? For code clearness and all that

test/stats/stats_test_utils.py
```diff
@@ -98,3 +101,15 @@ def buy_cooldown(self, last_trade: Trade) -> int:
...
+    return datetime.fromtimestamp(timestamp / 1000)
+
+
+def create_test_date(year=1970, month=1, day=1, hour=0, minute=0, second=0) -> datetime:
```
WillemGerritzen, line 110: Not sure I understand the purpose of the functions

test/utils/signal_frame.py
```diff
@@ -24,7 +26,7 @@ class MockOHLCVWithSignal(dict, object):
     def __init__(self, key):
         super().__init__()
         self.__key = key
-        self.current_time = 1577833200000  # 2020-01-01
+        self.current_time = ccxt.binance.parse8601("2020-01-01T00:00:00Z")
```
WillemGerritzen, line 29: Why don't we use datetime.timestamp(year=2020, month=1, day=1) instead? Would not require  ccxt and would be more inline with how we handle dates throughout the test IMO

WillemGerritzen (changes requested):
Left some comments, maybe worth talking offline for some of them?

modules/stats/drawdown/for_portfolio.py
```diff
@@ -37,7 +38,7 @@ def get_max_realised_drawdown_for_portfolio(realised_profits_per_timestamp: dict
     return max_realised_drawdown - 1
 
 
-def get_longest_drawdown(per_timestamp_dict: dict) -> dict[str, timedelta | bool]:
+def get_longest_drawdown(per_timestamp_dict: dict) -> Dict[str, timedelta or bool]:
```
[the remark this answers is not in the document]
sheridavandenbent: Fair, but tbf I only changed this so I could run the code on my side - this is part of the longest drawdown code, so I'd say do change it when you make a PR for the longest drawdown addition. But before you do that, I'd like to talk with you and see if we can figure out a way to not have 'or' or 'union' functionality in the return types. I'd personally rather have timedelta or 'none' for example, but I'm not sure if that would work for you

modules/stats/metrics/profit_ratio.py
sheridavandenbent: That's a good kind of nitpicking! Good idea! I'll change it

sheridavandenbent: Done!

modules/stats/tradingmodule.py
sheridavandenbent: Because we need to save the starting capital somewhere - we can't save it on the starting timestep because it will be overwritten the moment there's a buy signal on the first tick, since it'll buy, and remove the fee from the capital.

modules/stats/utils.py
sheridavandenbent: Yes, good idea. :v:

resources/setup/strategies/my_strategy.py
sheridavandenbent: Errrrr.. Good catch! 😅

resources/setup/strategies/my_strategy_advanced.py
```diff
@@ -20,7 +20,7 @@ def additional_pairs():
         :return: list
         """
         return [
-            ("BTC/USDT", "4h")
+            ("BTC/USDT", "1d")
```
sheridavandenbent, line 23:
```suggestion
            ("BTC/USDT", "4h")
```

resources/setup/strategies/my_strategy_advanced.py
```diff
@@ -38,7 +38,7 @@ def generate_indicators(self, dataframe: DataFrame, additional_pairs=None) -> Da
         add_btc_data['rsi'] = ta.RSI(add_btc_data, timeperiod=14)
         # Finally, merge the additional btc data with the standard dataframe.
         # The new column will have the following format: ['<column>_<pair>_<timeframe>']
-        dataframe = self.join_additional_data(dataframe, add_btc_data, self.timeframe, "4h")
+        dataframe = self.join_additional_data(dataframe, add_btc_data, self.timeframe, "1d")
```
sheridavandenbent, line 41:
```suggestion
        dataframe = self.join_additional_data(dataframe, add_btc_data, self.timeframe, "4h")
```

resources/setup/strategies/my_strategy_advanced.py
```diff
@@ -103,7 +103,7 @@ def buy_signal(self, dataframe: DataFrame) -> DataFrame:
                     (dataframe['fastd'] > dataframe['fastk']) &
                     (dataframe['fastd'] > 0)
                 ) &
-                (dataframe['rsi_BTC/USDT_4h'] < 30)     # additional pair usage
+                (dataframe['rsi_BTC/USDT_1d'] < 30)     # additional pair usage
```
sheridavandenbent, line 106:
```suggestion
                (dataframe['rsi_BTC/USDT_4h'] < 30)     # additional pair usage
```

test/stats/stats_test_utils.py
sheridavandenbent: Yes, because the tests are written in such a way that 2020 was the base point. I can change the tests to have the test config reflect the default config value, but this was faster

test/stats/stats_test_utils.py
sheridavandenbent: Done, and cleaned up the pandas import too

test/stats/stats_test_utils.py
sheridavandenbent: Yes good point, thanks

test/stats/stats_test_utils.py
sheridavandenbent: They're there to make it easy to see in tests which dates are compared - instead of creating the timestamp, you'd call the create_test_timestamp with the correct year, month, day, etc. That way it's clear to see immediately which date and time it should be, if the test fails.

sheridavandenbent: I just think it looks cleaner in the tests now :)

test/utils/signal_frame.py
sheridavandenbent: I've changed it to use the new test date creator I made!

utils/utils.py
```diff
@@ -79,6 +79,8 @@ def parse_timeframe(timeframe_str: str):
         timeframe_time = int(items[1]) * MINUTE
     elif items[2] == 'h':
         timeframe_time = int(items[1]) * HOUR
+    elif items[2] == 'd':
+        timeframe_time = int(items[1]) * DAY
```
sheridavandenbent, line 83:
```suggestion
```

sheridavandenbent:
@WillemGerritzen I've gone through your comments and implemented them and/or commented on them! This PR is ready for re-reviewing!

WillemGerritzen (approved):
lgtm

PR #528, opened 2022-02-21
[ENG-90] Add test for default config file

WillemGerritzen:
# What has been changed?

Adds a test that can be run before committing to make sure the config file doesn't stray from its default values


sheridavandenbent (changes requested):
Looks great, good idea. Two points of feedback:

1. It's currently in the test/stats folder, but it's not testing the stats - I'd put it in the test/setup/config folder, personally.
2. There are two more config files that should be tested and should be equal at all time: resources/setup/config-defaults.json, and resources/setup/strategies/config.json - the first one is used for adding new values if there are values missing from people's already existing config, and the second one is the base for the executable. I see that there's a small discrepancy between that one and the 'develop' config that we use, so I'll fix that straight away in a new PR. But it would be nice for this test to include those two as well (as obviously it's easy for discrepancies to show up there). Do you think it's doable to include those two files?

config.json
```diff
@@ -23,7 +23,7 @@
   "currency": "USDT",
   "fee": 0.25,
   "strategy-name": "MyStrategy",
-  "strategies-folder": "resources/setup/strategies",
+  "strategies-folder": "strategies",
```
sheridavandenbent, line 26: Why is this part deleted? This config file is used when running the engine via python, and therefore the strategies will be in resources/setup/strategies, no?

main_controller.py
```diff
@@ -39,11 +39,14 @@ def objective(trial: Trial):
...
         except KeyboardInterrupt:
             print_info("Quitting hyperoptimization.")
+
             try:
                 print_info(f"Best results: {study.best_params}")
+
             except ValueError:
                 print_info("No trials completed yet. Results not available.")
-            sys.exit()
+
```
sheridavandenbent, line 51: Why is this line removed?

sheridavandenbent: Since if you remove it, it will run the code underneath, and show an error because the best results can't be retrieved

sheridavandenbent (changes requested):
I've left a few comments regarding the placement of the 'strategies' folder - when running the engine with python, the strategies are in /resources/setup/strategies, but when running the engine with the executable they're in /strategies.

sheridavandenbent (approved):
I like your solution to the strategy file placement! Looks great. Approved! ✅

PR #530, opened 2022-02-22
[ENG 59] Add engine use metrics

sheridavandenbent:
# Results of merging this
Adds the engine use metrics functionality. The engine can be run anonymously by adding the '-nostats True' argument, when running with python or running with the executable.

Developers should send their personal uuid (to be found in ./statistics_id.txt') to me so I can put you on the developer list, and include you in the developer statistics and not include you in the user statistics. Do copy your uuid txt file to your executable folder if you have one.

This functionality works by making a post request to a handler that we have running on Scaleway. To change the way the handler works (for example, to add new fields to the statistics), see the engine-metrics github project. The creation of pretty graphs and tables can also be done in the jupyter notebook in the same github project.


.gitignore
```diff
@@ -134,9 +134,11 @@ dmypy.json
 # .DS_Store
 .DS_Store
 
-**/.idea/**
+**/.idea/**git staqgit
```
sheridavandenbent, line 137:
```suggestion
**/.idea/**
```

config.json
```diff
@@ -24,7 +24,7 @@
   "fee": 0.25,
   "strategy-name": "MyStrategy",
   "strategies-folder": "resources/setup/strategies",
-  "plots": true,
+  "plots": false,
```
sheridavandenbent, line 27:
```suggestion
  "plots": true,
```

requirements.txt
```diff
@@ -16,4 +16,5 @@ openpyxl==3.0.9
 certifi==2021.10.8
 pandas_ta==0.3.14b
 scipy==1.7.3
-pytz==2021.3
\ No newline at end of file
+requests-futures==1.0.0
+pytz==2021.3
```
WillemGerritzen, line 20: In all three requirements file an EOF line was added. Clearly, this is such a good PR I'm left to point this kind of silly thing out but is it on purpose?

WillemGerritzen (changes requested):
made 1 comment

requirements.txt
sheridavandenbent: Fixed it!

WillemGerritzen (approved):
nice!

PR #534, opened 2022-02-25
[ENG-96] Fix default config tests

WillemGerritzen:
# What has been changed?

Some elements of the default configs turned out not to be tested. This PR fixes that.


sheridavandenbent (approved):
👍

PR #535, opened 2022-02-25
[ENG-95] Disable plots via cli parameter

sheridavandenbent:
# Results of merging this
Before, the plotting behaviour could not be influenced by using a cli parameter. This PR fixes that.


WillemGerritzen (changes requested):
Ignore this, it's a mistake

WillemGerritzen (approved):
lgtm!

PR #536, opened 2022-02-28
[ENG-97] Add resources flag to executable

WillemGerritzen:
# What has been changed?

Adds the possibility of specifying the location of the directory containing the config and strats when running the engine as an executable. Allows for easier scripting and to run the engine from a different directory.

### DIFF
--- a/main.py
+++ b/main.py
@@ -10,9 +10,9 @@ from cli.arg_parse import execute_for_args
 from cli.checks.engine_use_statistics import engine_statistics
 from cli.checks.latest_version import print_warning_if_version_outdated
 from cli.prepare_workspace import prepare_workspace
-from cli.print_utils import print_debug, is_verbosity, print_warning
+from cli.print_utils import print_debug, is_verbosity
 from main_controller import MainController
-from utils.utils import check_internet_connection
+from utils.utils import check_internet_connection, prepend_resource_dir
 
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -27,9 +28,6 @@
 def main(online: bool):
     if online:
         print_warning_if_version_outdated()
-    else:
-        print_warning("You are not connected to the internet. Certain functionality (like downloading new candle data)"
-                      " will not be available.")
```
sheridavandenbent, line 32: Why remove this warning?

resources/specification.json
```diff
@@ -124,9 +124,12 @@
     }
   },
   {
-    "name": "strategies-folder",
+    "name": "strategies_folder",
```
sheridavandenbent, line 127: Why write it with an underscore? Everything in the specification.json (except export_results, which should be changed imho) is written with a dash instead

sheridavandenbent: Furthermore it's not changed in the config files either (also a dash there) so the specification and configs are now out of sync

sheridavandenbent (changes requested):
Works great! I have left two comments on the code.

main.py
WillemGerritzen: There's already a warning upstream in the code.

sheridavandenbent: Very valid reason 👍

sheridavandenbent (approved):
Looks great!

